PR #55, opened 2024-05-02
Use CO2.js for Carbon Intensity values

mgriffin-scottlogic:
Also updated our WorldLocation to use the same format as CO2.js
Display the list in assumptions as grams instead of kilograms as it's easier to read and more commonly used form.
Updated some tests due to minor variations.

### DIFF
--- a/src/app/assumptions-and-limitation/assumptions-and-limitation.component.html
+++ b/src/app/assumptions-and-limitation/assumptions-and-limitation.component.html
@@ -92,21 +92,21 @@
   <h4 class="text-lg">Carbon Intensity</h4>
   <p class="text-sm">
     We make use of the latest available Carbon Intensity figures from the
-    <a class="underline" href="https://ember-climate.org/data/data-tools/data-explorer/">Ember Data Explorer</a> for the
-    regions we make available. At present these are:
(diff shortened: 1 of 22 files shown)


src/app/assumptions-and-limitation/assumptions-and-limitation.component.html
```diff
@@ -95,23 +95,25 @@ <h4 class="tce-text-lg">Power consumption</h4>
   <h4 class="tce-text-lg">Carbon Intensity</h4>
   <p class="tce-text-sm">
     We make use of the latest available Carbon Intensity figures from the
-    <a class="tce-underline" href="https://ember-climate.org/data/data-tools/data-explorer/">Ember Data Explorer</a> for
-    the regions we make available. At present these are:
+    <a class="tce-underline" href="https://ember-climate.org/data/data-tools/data-explorer/">Ember Data Explorer</a> via
+    the CO2.js library. We limit the range of regions that can be selected, which are currently:
```
jantoun-scottlogic, line 99: Is it worth adding a link to the CO2.js library here? Also, not something we can do anything about but it's taking a long time to connect to the Ember website for me at the moment (around 30 seconds). My latest attempt to connect (via Google) actually timed out altogether. Might just be a temporary thing but maybe worth keeping an eye on.

jantoun-scottlogic (approved):

PR #57, opened 2024-05-09
Make tooltips toggle option

jantoun-scottlogic:
Makes question clarification tooltips a toggle option rather than a popup


src/app/expansion-panel/expansion-panel.component.ts
```diff
@@ -8,12 +9,23 @@ import { ChangeDetectionStrategy, Component, EventEmitter, Output } from '@angul
   templateUrl: './expansion-panel.component.html',
   changeDetection: ChangeDetectionStrategy.OnPush,
 })
-export class ExpansionPanelComponent {
+export class ExpansionPanelComponent implements OnInit {
+  @Input() questionPanel?: boolean = false;
```
mgriffin-scottlogic, line 13:
```suggestion
  @Input() configOverride?: Partial<ExpansionPanelConfig>;
```
This would allow users of the component to customise as much as they want

src/app/expansion-panel/expansion-panel.component.ts
```diff
@@ -8,12 +9,23 @@ import { ChangeDetectionStrategy, Component, EventEmitter, Output } from '@angul
...
   @Output() expandedChanged = new EventEmitter<void>();
-  public expanded: boolean = true;
+  public expanded: boolean = defaultConfig.startsExpanded;
```
mgriffin-scottlogic, line 15:
```suggestion
  public config: ExpansionPanelConfig = defaultConfig;
```
Combine all of the public parts of the config.

src/app/expansion-panel/expansion-panel.component.ts
```diff
@@ -8,12 +9,23 @@ import { ChangeDetectionStrategy, Component, EventEmitter, Output } from '@angul
...
+  public contentContainerStyles: string = defaultConfig.contentContainerStyles;
 
   toggle() {
     this.expanded = !this.expanded;
     this.expandedChanged.emit();
   }
+
+  ngOnInit(): void {
+    if (this.questionPanel) {
+      this.expanded = questionPanelConfig.startsExpanded;
```
mgriffin-scottlogic, line 26:
```suggestion
      this.config = {
        ...this.config,
        ...this.configOverride,
      };
```
This should give a config with only the requested parts changed, without having to update if we add more config options.

src/app/expansion-panel/expansion-panel.constants.ts
```diff
@@ -0,0 +1,17 @@
...
+  contentContainerStyles: string;
+};
+
+export const defaultConfig: ExpansionPanelConfig = {
+  startsExpanded: true,
+  buttonStyles: 'material-icons-outlined tce-text tce-text-slate-600 hover:tce-bg-slate-200 hover:tce-rounded',
+  contentContainerStyles: '',
+};
+
+export const questionPanelConfig: ExpansionPanelConfig = {
```
mgriffin-scottlogic, line 13: This should live on the carbon-estimator-form for now, passing it in to the configOverride input when used for help display.

mgriffin-scottlogic (changes requested):
Code changes may not work 100% but hope that gives an idea of what I was thinking?

src/app/expansion-panel/expansion-panel.component.ts
mgriffin-scottlogic: I've also seen the non-null assertion used in that way, one of those things that's technically unsafe but you'd find out pretty quickly if it was ever changed so it isn't set.

Another option would be to rename startsExpanded to just expanded and use/toggle the value that's inside the config object. I don't know if either feels better to you?

src/app/expansion-panel/expansion-panel.component.ts
jantoun-scottlogic: Done :)

src/app/expansion-panel/expansion-panel.constants.ts
jantoun-scottlogic: I've moved this to the carbon-estimator-form. Let me know if it's not in quite the right place/files.

src/app/expansion-panel/expansion-panel.component.ts
mgriffin-scottlogic: Cool, that's fine by me to go for the non-null assertion version.

The point about local state and not modifying things that are passed in feels valid. I think in this instance it would be okay to use the expanded value as the ... spread operator creates a copy of the bool property but if we had some more complex nested objects then you would potentially be modifying external objects.

jantoun-scottlogic: Ah yes, I hadn't considered that we were copying into a new object. I still feel like the config object that gets passed in should be immutable but the internal config would need a mutable 'expanded' field.

jantoun-scottlogic: I think overall I prefer keeping the state and the config separate and if we added more state in the future it wouldn't necessarily belong in the config object so we could end up with state stored in slightly different ways in the component. Happy to defer though if you prefer having it as a config field.

docs/components.md
```diff
@@ -13,8 +13,6 @@ flowchart TB
     carbon-estimation`"]
     Assumptions["`AssumptionsAndLimitationComponent
     assumptions-and-limitation`"]
-    HelperInfo["`HelperInfoComponent
```
mgriffin-scottlogic, line 16: Just realised, you'll also need to remove the reference to HelperInfo later in this diagram (`CarbonEstimatorForm --> HelperInfo & Note`). With it there it creates a box and connection to it still.

jantoun-scottlogic: Ah thanks! I've removed the description of the component from the docs as well.

mgriffin-scottlogic (approved):

PR #58, opened 2024-05-17
Ensure that form can always be submitted when opting out of end user calculations

mgriffin-scottlogic:
I think this is the best way to achieve this, found the following blog covering some options: https://netbasal.com/three-ways-to-dynamically-alter-your-form-validation-in-angular-e5fd15f1e946

1. Dynamically Add a Control - This seems like it could cause problems converting to `EstimatorValues` in `handleSubmit()`.
2. Dynamically Add Validators - We don't create the `MinValidator` explicitly so we can't remove it (the `min` in the `<input>` creates one automatically).
3. Disable the Control - Basically gives us what we want and works no matter how many validators the control has.

I also noticed that it was happy to submit when the value was deleted completely, the `required` validator prevents this. And even though we don't use the `monthlyActiveUsers` value when `noDownstream` is set, I thought it was safer to still make it zero rather than null in this case.

### DIFF
--- a/src/app/carbon-estimator-form/carbon-estimator-form.component.ts
+++ b/src/app/carbon-estimator-form/carbon-estimator-form.component.ts
@@ -126,9 +126,15 @@ export class CarbonEstimatorFormComponent implements OnInit {
       this.changeDetector.detectChanges();
     });
 
-    this.estimatorForm
-      .get('downstream.noDownstream')
-      ?.valueChanges.subscribe(noDownstream => (this.noDownstream = noDownstream));
+    this.estimatorForm.get('downstream.noDownstream')?.valueChanges.subscribe(noDownstream => {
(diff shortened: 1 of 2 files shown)


src/app/carbon-estimator-form/carbon-estimator-form.component.ts
jantoun-scottlogic, line 1: When the number of users is set to 0, the `Calculate` button appears enabled when the section is expanded again. (This doesn't happen when the number of users field is blank.) It's not until you click somewhere on the page e.g. try to submit that the button becomes disabled.
calculate-button-enabled-invalid-form

mgriffin-scottlogic: Good spot, I'm not 100% why this doesn't get picked up but we do a similar thing for the cloud services checkbox above, which seems to fix it.

jantoun-scottlogic (approved):

PR #59, opened 2024-05-20
SFD-78: Add debounce to chart resize

jantoun-scottlogic:
Jira ticket: [SFD-78](https://scottlogic.atlassian.net/jira/software/projects/SFD/boards/104?selectedIssue=SFD-78)

Adds debounce to chart resize.


mgriffin-scottlogic (approved):
Looks good to me. It's generally recommended not to explicitly unsubscribe from subscriptions but I think this is one of those cases where it's unavoidable.

PR #60, opened 2024-05-21
SFD-89 Create merge pipeline

jantoun-scottlogic:
Jira ticket: [https://scottlogic.atlassian.net/browse/SFD-89](https://scottlogic.atlassian.net/browse/SFD-89)

Adds a workflow that runs when a pull request is opened and when new commits are pushed to an open pull request. The workflow checks the app builds successfully and runs the unit tests.

### DIFF
--- a/.github/actions/set-up-environment/action.yml
+++ b/.github/actions/set-up-environment/action.yml
@@ -1,0 +1,17 @@
+name: 'Set up environment'
+description: 'Sets up Node and installs dependencies'
+runs:
+  using: 'composite'
+  steps:
+    - name: Set up Node
+      uses: actions/setup-node@v4
+      with:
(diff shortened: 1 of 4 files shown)


jantoun-scottlogic:
Just added some documentation about how to skip running a workflow when pushing to a PR branch. It looks like this may actually bypass the requirement for the workflow to pass against the latest commit to allow merging so it's possible we may need to change a setting somewhere if we want to enforce that.
 
There are some other things we could tweak as well although we don't necessarily need to do them in this PR. E.g. it would be good to exclude some changes from triggering the pipeline/requiring it to pass before merging like if someone just updates the docs.

mgriffin-scottlogic (approved):

PR #71, opened 2024-06-10
SFD-97: Add accessibility to disabled 'Calculate' Button

jantoun-scottlogic:
[SFD-97 Jira ticket](https://scottlogic.atlassian.net/browse/SFD-97)

Makes the 'Calculate' button focusable when disabled.

### DIFF
--- a/src/app/carbon-estimator-form/carbon-estimator-form.component.html
+++ b/src/app/carbon-estimator-form/carbon-estimator-form.component.html
@@ -235,10 +235,8 @@
 
   <div class="tce-flex tce-gap-4 tce-flex-row-reverse">
     <button
-      class="{{
-        'tce-button-calculate tce-px-3 tce-py-2' +
-          (!estimatorForm.valid ? ' tce-opacity-50 tce-cursor-not-allowed' : '')
-      }}"
+      class="tce-button-calculate tce-px-3 tce-py-2"
+      [ngClass]="!estimatorForm.valid ? 'tce-opacity-50 tce-cursor-not-allowed' : ''"
       type="submit"
       [attr.aria-disabled]="!estimatorForm.valid">
       Calculate
--- a/src/app/carbon-estimator-form/carbon-estimator-form.component.html
+++ b/src/app/carbon-estimator-form/carbon-estimator-form.component.html
@@ -1,4 +1,4 @@
-<form (ngSubmit)="handleSubmit()" [formGroup]="estimatorForm" class="tce-w-full tce-flex tce-flex-col tce-gap-6">
+<form [formGroup]="estimatorForm" class="tce-w-full tce-flex tce-flex-col tce-gap-6">
   <div formGroupName="upstream">
     <ng-container *ngTemplateOutlet="sectionHeader; context: formContext.upstream"></ng-container>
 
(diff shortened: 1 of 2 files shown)


src/app/carbon-estimator-form/carbon-estimator-form.component.html
```diff
@@ -1,4 +1,4 @@
-<form (ngSubmit)="handleSubmit()" [formGroup]="estimatorForm" class="tce-w-full tce-flex tce-flex-col tce-gap-6">
```
jantoun-scottlogic, line 1: I couldn't find a nice way to make the "submit" type "Calculate" button trigger conditionally based on whether the form was valid so I changed it to type="button" and set handleSubmit as the click handler with some extra logic so it would only submit the form when it was valid.

mgriffin-scottlogic: This means that you don't get the diagram updating when you hit enter on a text form, which I think would be nice to keep. Is there anything preventing us from keeping this as the submit handler and just have the same early out on whether the form is valid?

src/app/carbon-estimator-form/carbon-estimator-form.component.html
```diff
@@ -235,9 +235,13 @@
 
   <div class="tce-flex tce-gap-4 tce-flex-row-reverse">
     <button
-      class="tce-button-calculate tce-px-3 tce-py-2 disabled:tce-opacity-50 disabled:tce-cursor-not-allowed"
-      type="submit"
-      [disabled]="!estimatorForm.valid">
+      class="{{
```
mgriffin-scottlogic, line 238: Might be cleaner to use a combination of class and ngClass for the conditional parts.

src/app/carbon-estimator-form/carbon-estimator-form.component.html
jantoun-scottlogic: Yes, we can do that. We may need to find a way to provide feedback in this scenario when the form is invalid and hence not submitted when pressing the enter key. Otherwise it could be confusing if a user is working on a valid portion of the form but hitting enter doesn't do anything because another part of the form is invalid.

jantoun-scottlogic: I've reverted to using ngSubmit to trigger the handler so pressing enter triggers submission provided the form is valid.

src/app/carbon-estimator-form/carbon-estimator-form.component.html
jantoun-scottlogic: Done :)

src/app/carbon-estimator-form/carbon-estimator-form.component.html
```diff
@@ -235,9 +235,10 @@
 
   <div class="tce-flex tce-gap-4 tce-flex-row-reverse">
     <button
-      class="tce-button-calculate tce-px-3 tce-py-2 disabled:tce-opacity-50 disabled:tce-cursor-not-allowed"
+      class="tce-button-calculate tce-px-3 tce-py-2"
+      [ngClass]="!estimatorForm.valid ? 'tce-opacity-50 tce-cursor-not-allowed' : ''"
```
mgriffin-scottlogic, line 239:
```suggestion
      [ngClass]="{ 'tce-opacity-50 tce-cursor-not-allowed': !estimatorForm.valid }"
```
Just a minor thing, this way avoids the ternary/empty string.

jantoun-scottlogic: Ah ok, thanks! I've made that change.

src/app/carbon-estimator-form/carbon-estimator-form.component.html
mgriffin-scottlogic: > We may need to find a way to provide feedback in this scenario when the form is invalid and hence not submitted when pressing the enter key.
Maybe make another spike around that? I'm not sure what's recommended and it is a bit inconsistent when Enter is going to submit the form vs select an option from a dropdown with keyboard controls etc. Was surprised that Enter submits the form when you have a checkbox focussed.

mgriffin-scottlogic (approved):

mgriffin-scottlogic:
Approved, if you want to add a link to this PR to the ticket and move it to ready to test.

PR #76, opened 2024-06-12
SFD-107: Change main pipeline to not publish from every commit

jantoun-scottlogic:
[SFD-107 Jira ticket](https://scottlogic.atlassian.net/browse/SFD-107)

Removes the merge trigger for the pages deployment workflow leaving only the manual trigger.

### DIFF
--- a/.github/workflows/pages.yml
+++ b/.github/workflows/pages.yml
@@ -1,10 +1,6 @@
 name: Deploy estimator Pages
 
 on:
-  # Runs on pushes targeting the default branch
-  push:
-    branches: ['main']
-
   # Allows you to run this workflow manually from the Actions tab
   workflow_dispatch:
 


mgriffin-scottlogic (approved):
Approved, I think this can be merged in now?

I was going to suggest that workflow_dispatch should also be restricted to the main branch but apparently that's not a thing: https://stackoverflow.com/questions/72398993/github-action-workflow-dispatch-how-to-always-use-the-master-branch-remove-o

jantoun-scottlogic:
> Approved, I think this can be merged in now?
> 
> I was going to suggest that workflow_dispatch should also be restricted to the main branch but apparently that's not a thing: https://stackoverflow.com/questions/72398993/github-action-workflow-dispatch-how-to-always-use-the-master-branch-remove-o

We could add a step at the start to check if the branch is main and exit if not.

jantoun-scottlogic:
> > Approved, I think this can be merged in now?
> > I was going to suggest that workflow_dispatch should also be restricted to the main branch but apparently that's not a thing: https://stackoverflow.com/questions/72398993/github-action-workflow-dispatch-how-to-always-use-the-master-branch-remove-o
> 
> We could add a step at the start to check if the branch is main and exit if not.

As discussed, we'll leave it as it is for now so people can publish from other branches if desired and revisit if needed.

PR #77, opened 2024-06-12
Migrate playwright tests to repo

mgriffin-scottlogic:
Used git subtree command to bring @sdun-scottlogic's repo into playwright directory.

    git subtree add -P playwright https://github.com/sdun-scottlogic/sl-tech-carbon-estimator-tests.git HEAD

As far as I can understand this can be treated as a one time operation and we shouldn't need to depend on the source repo going forward.

There are now two new npm scripts, `playwright-install` and `playwright-test` to install the required python packages and run all tests respectively. This now also takes care of starting the server automatically and waiting for it to be ready before starting the tests.

I was thinking that it would be good to get this merged in soon so that we can start fixing tests within this repo.


playwright/tests-examples/demo-todo-app.spec.ts
jantoun-scottlogic, line 1: I think we can delete this file (and the tests-examples folder). The tests aren't related to our app and it's causing issues with the running the unit tests.

jantoun-scottlogic (changes requested):

playwright/tests-examples/demo-todo-app.spec.ts
mgriffin-scottlogic: Hehe, I've just literally figured out how to exclude the playwright folder from the test config, I thought it would be much easier than it was but probably right that we don't actually need it.

playwright/playwright-report/index.html
jantoun-scottlogic, line 1: Should this be in the .gitignore?

playwright/README.md
```diff
@@ -0,0 +1,84 @@
...
+If you would like to view test flow in browser as it runs step by step but find that tests run too quick (browser window opens and closes too quickly) you can slow test down by adding `--slowmo`
+
+Example: `pytest test_2_EstTool_HappyPath_Default.py --headed --slowmo 1000` (or 2000)
+
+Tests can be run in different browsers (Use `chromium` for Chrome, `firefox` for Firefox & `webkit` for Safari). The following test will run on firefox and browser window will be open.
+
+Example: `pytest test_2_EstTool_HappyPath_Default.py --browser=firefox --headed`
+
+Note: All tests are pointing to `http://localhost:4200/`, if there is an issue with this (possibly 4200 may already be in use) then you may be directed to a different url, example `http://localhost:5600/`
+If this is the case then then the code within each individual test to be run may need updated to match with localhost address otherwise the test(s) will fail. No global file was available to cover this scenario.
```
jantoun-scottlogic, line 68:
```suggestion
If this is the case then then the code within each individual test to be run may need to be updated to match with localhost address otherwise the test(s) will fail. No global file was available to cover this scenario.
```

jantoun-scottlogic: I'm not sure updating the tests would be what we want to do here because that would cause problems for users whose app is running on localhost:4200. We could sort this out after merging though.

playwright/.github/workflows/playwright.yml
```diff
@@ -0,0 +1,27 @@
+name: Playwright Tests
+on:
+  push:
+    branches: [ main, master ]
```
jantoun-scottlogic, line 4:
```suggestion
    branches: [ main ]
```

playwright/.github/workflows/playwright.yml
```diff
@@ -0,0 +1,27 @@
...
+  pull_request:
+    branches: [ main, master ]
```
jantoun-scottlogic, line 6:
```suggestion
    branches: [ main ]
```

playwright/.github/workflows/playwright.yml
```diff
@@ -0,0 +1,27 @@
...
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v4
+    - uses: actions/setup-node@v4
+      with:
+        node-version: lts/*
+    - name: Install dependencies
+      run: npm ci
+    - name: Install Playwright Browsers
+      run: npx playwright install --with-deps
```
jantoun-scottlogic, line 19: Will these commands work with the Python version of Playwright?

playwright/.github/workflows/playwright.yml
jantoun-scottlogic, line 1: I'm wondering if it would be better to set up the Playwright workflow after merging in and fixing the the tests. I think we will probably need to modify this to use the Playwright Python commands and we'll need to add a command to start the app as well.

jantoun-scottlogic: [This](https://playwright.dev/python/docs/ci-intro) looks like what we want.

jantoun-scottlogic: I realised we don't need an extra command to start the app since it's taken care of in estimator_server.py.

playwright/README.md
mgriffin-scottlogic: Yeah, there's a better way of handling this that we might want to get around to as another task, you can pass --base-url on the commandline and then just do `page.goto("/")` in the tests. Would just need to pick up on the url in the estimator_server fixture to make sure it doesn't try to start the server if a different address has been specified.

playwright/.github/workflows/playwright.yml
mgriffin-scottlogic: Leaving for a later task

playwright/.github/workflows/playwright.yml
mgriffin-scottlogic: Leaving for a later task

playwright/.github/workflows/playwright.yml
jantoun-scottlogic: Yeah, probably easiest to sort after the tests are merged, updated and passing.

jantoun-scottlogic (approved):

PR #78, opened 2024-06-13
SFD-124: Limit the size/shape that the diagram can be

jantoun-scottlogic:
[SFD-124 Jira ticket](https://scottlogic.atlassian.net/browse/SFD-124)

Caps the chart height based on screen height to prevent issues with the chart becoming stretched when the component is displayed in a tall iFrame.

In iFrame before fix:
in-iframe-without-fix

In iFrame after fix:
in-iframe-with-fix

I've capped based on screen height rather than aspect ratio for the time being, mostly due to the screen height cap seeming to work ok and being unfamiliar with using aspect ratio to set height and how it would work in various contexts. I think it looks ok on mobile views (checked with the Chrome device toggle) and laptop but we might need to revisit and use aspect ratio if we aren't getting the desired effect in some settings.

### DIFF
--- a/src/app/carbon-estimation/carbon-estimation.component.spec.ts
+++ b/src/app/carbon-estimation/carbon-estimation.component.spec.ts
@@ -72,7 +72,7 @@ describe('CarbonEstimationComponent', () => {
     component.onResize(2000, 1000, 2000);
 
     expect(component.chart?.updateOptions).toHaveBeenCalledOnceWith({
-      chart: { height: 1400 }, // Height will be capped at a percentage of the screen height
+      chart: { height: 1500 }, // Height will be capped at a percentage of the screen height
     });
   });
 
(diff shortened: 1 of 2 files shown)


mgriffin-scottlogic (changes requested):
This looks to work fine everywhere, the 0.7 feels a little arbitrary so maybe we just put it into a constant for some explanation?

As for aspect ratio, I'm not sure that there's anything that deals with it automatically, would just have to calculate from the width ie.
```
const MIN_RATIO = 9 / 16;
const MAX_RATIO = 16 / 9;
...
const minHeight = MIN_RATIO * innerWidth;
const maxHeight = MAX_RATIO * innerWidth;
```
Or something like that.

mgriffin-scottlogic (dismissed):

mgriffin-scottlogic (approved):

PR #85, opened 2024-06-18
SFD-98: Chart scaling on zoom bugfix

jantoun-scottlogic:
[SFD-98 Jira ticket](https://scottlogic.atlassian.net/jira/software/projects/SFD/boards/104)

Scales the chart size with the browser zoom factor to fix a bug where the chart would become squashed when the browser zoom was increased.

Before fix:
before-fix-500-percent-zoom

After fix:
after-fix-500-percent-zoom

### DIFF
--- a/src/app/carbon-estimation/carbon-estimation.component.spec.ts
+++ b/src/app/carbon-estimation/carbon-estimation.component.spec.ts
@@ -48,11 +48,16 @@ describe('CarbonEstimationComponent', () => {
   });
 
   it('should set chart height when inner height is more than base height', () => {
+    console.log(`window.innerHeight: ${window.innerHeight}`);
     spyOnProperty(window, 'innerHeight').and.returnValue(1000);
     spyOnProperty(component.detailsPanel.nativeElement, 'clientHeight').and.returnValue(200);
 
     component.ngOnInit();
     fixture.detectChanges();
+
+    console.log(`window.innerHeight: ${window.innerHeight}`);
+    console.log(`screen height: ${window.screen.height}`);
+    console.log(`device pixel ratio: ${window.devicePixelRatio}`);
 
     expect(component.chartOptions.chart.height).toBe(1000 - estimatorBaseHeight - 200);
   });
--- a/src/app/carbon-estimation/carbon-estimation.component.spec.ts
+++ b/src/app/carbon-estimation/carbon-estimation.component.spec.ts
@@ -87,14 +87,16 @@ describe('CarbonEstimationComponent', () => {
     });
   });
 
-  it('should scale chart height according to browser zoom factor', () => {
-    spyOn(component.chart as ChartComponent, 'updateOptions');
-    spyOnProperty(component.detailsPanel.nativeElement, 'clientHeight').and.returnValue(200);
+  [0.5, 1, 2, 5].forEach(browserZoomFactor => {
+    it(`should scale chart height appropriately if the browser zoom factor is ${browserZoomFactor}`, () => {
+      spyOn(component.chart as ChartComponent, 'updateOptions');
+      spyOnProperty(component.detailsPanel.nativeElement, 'clientHeight').and.returnValue(200);
 
-    component.onResize(1500, 1000, 2000, 2);
+      component.onResize(1500, 1000, 2000, browserZoomFactor);
 
-    expect(component.chart?.updateOptions).toHaveBeenCalledOnceWith({
-      chart: { height: (1500 - estimatorBaseHeight - 200) * 2 },
+      expect(component.chart?.updateOptions).toHaveBeenCalledOnceWith({
+        chart: { height: (1500 - estimatorBaseHeight - 200) * browserZoomFactor },
+      });
     });
   });
 
--- a/src/app/carbon-estimation/carbon-estimation.component.spec.ts
+++ b/src/app/carbon-estimation/carbon-estimation.component.spec.ts
@@ -65,30 +65,8 @@ describe('CarbonEstimationComponent', () => {
     expect(component.chartOptions.chart.height).toBe(600 - estimatorBaseHeight - 200 - 100);
   });
 
-  it('should recalculate chart height on window resize, for laptop screen', () => {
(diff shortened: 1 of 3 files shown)


src/app/carbon-estimation/carbon-estimation.component.spec.ts
```diff
@@ -65,25 +65,53 @@ describe('CarbonEstimationComponent', () => {
     expect(component.chartOptions.chart.height).toBe(600 - estimatorBaseHeight - 200 - 100);
   });
 
-  it('should recalculate chart height on window resize, for laptop screen', () => {
+  [0.5, 1, 2, 5].forEach(browserZoomFactor => {
```
jantoun-scottlogic, line 68: Jasmine doesn't seem to have built in support for parameterised tests. This solution was suggested on Stack Overflow and in a few Medium posts.

mgriffin-scottlogic:
Just from a quick test of this myself, I'm still seeing an issue when it hits 400% zoom - I don't know if it's specific to my monitor resolution and scaling? (1920 x 1080 at 100%)

Haven't reviewed the code but I wonder if it could be falling prey to the opposite max height check if that's not also scaled with the zoom factor. There's a definite snap back to a small size as you get to 400:

### 300%

image

### 400%

image

mgriffin-scottlogic:
And strangely it seems fine if I scroll to 500% very quickly, I assume the debounce means the 400% state is missed. But if I roll back and forth at that point then 500% will be broken too.

jantoun-scottlogic:
> Just from a quick test of this myself, I'm still seeing an issue when it hits 400% zoom - I don't know if it's specific to my monitor resolution and scaling? (1920 x 1080 at 100%)
> 
> Haven't reviewed the code but I wonder if it could be falling prey to the opposite max height check if that's not also scaled with the zoom factor. There's a definite snap back to a small size as you get to 400:
> 
> ### 300%
> image
> 
> ### 400%
> image

Ok thanks, I'll take a look. I was getting different degrees of 'squashedness' of the same zoom factor when trying to reproduce the bug so there's definitely something going on that I haven't pinned down.

jantoun-scottlogic:
It looks like it's not scaling the chart if the screen is already zoomed when calculate is pressed either.

src/app/carbon-estimation/carbon-estimation.component.spec.ts
```diff
@@ -48,31 +48,36 @@ describe('CarbonEstimationComponent', () => {
...
     fixture.componentRef.setInput('extraHeight', '100');
     fixture.detectChanges();
     component.ngOnInit();
     fixture.detectChanges();
 
-    expect(component.chartOptions.chart.height).toBe(600 - estimatorBaseHeight - 200 - 100);
+    expect(component.chartOptions.chart.height).toBe(1000 - estimatorBaseHeight - 200 - 100);
   });
 
   it('should recalculate chart height on window resize, for laptop screen', () => {
```
jantoun-scottlogic, line 73: I separated out the tests that cap the height based on screen height. It's not directly related to the change in this PR but I had to update some of the chart height tests so I thought I'd update these as well while I was here.

src/app/carbon-estimation/carbon-estimation.component.ts
```diff
@@ -149,12 +149,23 @@ export class CarbonEstimationComponent implements OnInit, OnDestroy {
...
+    const calculatedHeight =
+      innerWidth < responsiveBreakpoint ?
+        innerHeight - this.estimatorBaseHeight - expansionPanelHeight + estimatorHeights.title
+      : innerHeight - this.estimatorBaseHeight - extraHeight - expansionPanelHeight;
+
+    // Cap on the mininum height of the chart to prevent the chart becoming squashed when zooming
+    // in on desktop browsers. (Zooming in results in window.innerHeight decreasing proportionally
+    // on most desktop browsers which can result in the calculatedHeight above becoming too small
+    // or even negative. N.B. Mobile browsers behave differently.)
+    const minChartHeight = 300;
```
jantoun-scottlogic, line 166: I was having some issues related to differences in behaviour between desktop and mobile browsers when trying to calculate a dynamic lower bound or rescaling the chart with zoom so I went with a fixed lower bound.

mgriffin-scottlogic: Given we limit the maximum height of the chart in `getChartHeight()`, it feels like it would be cleaner to do both clamping actions in there.

src/app/carbon-estimation/carbon-estimation.component.spec.ts
jantoun-scottlogic: I've removed the dependency on browserZoomFactor now so this no longer applies.

src/app/carbon-estimation/carbon-estimation.component.spec.ts
```diff
@@ -47,32 +47,26 @@ describe('CarbonEstimationComponent', () => {
...
     spyOnProperty(component.detailsPanel.nativeElement, 'clientHeight').and.returnValue(200);
-    component.ngOnInit();
-    fixture.detectChanges();
 
-    expect(component.chartOptions.chart.height).toBe(600 - estimatorBaseHeight - 200);
-  });
-
-  it('should set chart height to value if inner height more than base height plus extra height', () => {
-    spyOnProperty(component.detailsPanel.nativeElement, 'clientHeight').and.returnValue(200);
-    fixture.componentRef.setInput('extraHeight', '100');
```
mgriffin-scottlogic, line 60: Looks like we've lost all tests around setting the extraHeight input, I think we want to have at least one around this, even if it's changed a bit.

jantoun-scottlogic: I've added a test for the extraHeight input.

mgriffin-scottlogic (approved):

sdun-scottlogic:
Tested & Approved.
*** Note** - Playwright device emulator and browser Dev Tools to recreate multiple device behaviour - used as a best endeavours alternative for testing. May not fully or accurately replicate each device but is accepted as a reasonable alternative for these test cases.

PR #87, opened 2024-06-19
SFD-120 Add a disclaimer around use of the tool

mgriffin-scottlogic:
Shown above form

image

Also remains above Assumptions when shown

image

Spent a bit of time trying to work out why I couldn't add padding to the disclaimer element but haven't managed to narrow it down, so have just added to the internal expansion panel instead.

### DIFF
--- a/src/app/disclaimer/disclaimer.component.html
+++ b/src/app/disclaimer/disclaimer.component.html
@@ -1,9 +1,9 @@
-<expansion-panel class="tce-flex tce-justify-between tce-gap-2 tce-flex-col">
+<expansion-panel class="tce-flex tce-justify-between tce-gap-2 tce-flex-col tce-pb-4">
   <div header>
-    <h2 class="tce-text-2xl">Disclaimer</h2>
+    <h2 class="tce-text-xl">Disclaimer</h2>
   </div>
   <div content>
-    <p>
+    <p class="tce-text-sm">
       This tool is provided free of charge to assist you in identifying and measuring your carbon footprint, for
       information and guidance only. The scores and information provided should not be used for carbon reporting and are
       made available on an “as is” basis. Scott Logic Limited accepts no liability or responsibility for any use made by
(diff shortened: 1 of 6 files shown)


src/app/disclaimer/disclaimer.component.html
```diff
@@ -0,0 +1,15 @@
...
+  <div header>
+    <h2 class="tce-text-xl">Disclaimer</h2>
+  </div>
+  <div content>
+    <p class="tce-text-sm">
+      This tool is provided free of charge to assist you in identifying and measuring your carbon footprint, for
+      information and guidance only. The scores and information provided should not be used for carbon reporting and are
+      made available on an “as is” basis. Scott Logic Limited accepts no liability or responsibility for any use made by
+      any person or organisation of the information or calculations available on this website. Any reliance placed on
+      this website is taken at your own risk. You should read our full website terms and conditions before using this
```
jantoun-scottlogic, line 11: I think it would be good to have a link to the full terms and conditions here.

mgriffin-scottlogic: Would we link to the Scott Logic terms and conditions there? There was some confusion over this in the legal meeting as it could be used in two places, the TCS site doesn't have terms and conditions yet.

jantoun-scottlogic: I'm not sure but I think we need to check that and provide a link. Also, if we're publishing the tool as an NPM package then this could be on any website, not just one of ours so we need to consider that case as well. Is the idea that the disclaimer will still be on the site for people using this as an NPM package or will we just put a disclaimer in the README for users of the package?

mgriffin-scottlogic: Might need to look into this some more as www.scottlogic.com doesn't appear to have terms and conditions either.

mgriffin-scottlogic: Updated to link back to `TERMS.md`

jantoun-scottlogic (approved):

src/app/disclaimer/disclaimer.component.html
jantoun-scottlogic: Looks good!

PR #90, opened 2024-06-21
SFD-117 Some text needs appropriate styling for dark mode

mgriffin-scottlogic:
Have added additional html/css for the package build so that we can easily test this out without putting it into TCS. To use this version, start the server with: `ng serve --configuration npm-package`

image

When using `ng serve` on its own, it should still use the standard colours even with dark mode enabled. This is to match the Scott Logic site, which does not have any dark mode settings.

image

Mostly achieved by removing text colour specifications that aren't really needed. Added a colour for some elements that will always have a light background and some specific styling for the diagram legend text, which sets its own default colour and doesn't respond to dark mode.


src/app/carbon-estimation/carbon-estimation.component.css
```diff
@@ -37,4 +37,4 @@
 
 ::ng-deep .cell-tower-logo {
   background-image: url("data:image/svg+xml,%3Csvg xmlns='http://www.w3.org/2000/svg' viewBox='0 -960 960 960'%3E%3Cpath style='fill:white' d='M196-276q-57-60-86.5-133T80-560q0-78 29.5-151T196-844l48 48q-48 48-72 110.5T148-560q0 63 24 125.5T244-324l-48 48Zm96-96q-39-39-59.5-88T212-560q0-51 20.5-100t59.5-88l48 48q-30 27-45 64t-15 76q0 36 15 73t45 67l-48 48ZM280-80l135-405q-16-14-25.5-33t-9.5-42q0-42 29-71t71-29q42 0 71 29t29 71q0 23-9.5 42T545-485L680-80h-80l-26-80H387l-27 80h-80Zm133-160h134l-67-200-67 200Zm255-132-48-48q30-27 45-64t15-76q0-36-15-73t-45-67l48-48q39 39 58 88t22 100q0 51-20.5 100T668-372Zm96 96-48-48q48-48 72-110.5T812-560q0-63-24-125.5T716-796l48-48q57 60 86.5 133T880-560q0 78-28 151t-88 133Z'/%3E%3C/svg%3E");
-}
+}
```
jantoun-scottlogic, line 40: Looks like we just removed a newline from the end of this file. I think it should have one, Prettier seems to want to add one.

src/package-styles.css
```diff
@@ -0,0 +1,9 @@
+input, select {
+  @apply dark:tce-text-slate-600
+}
+
+@media (prefers-color-scheme: dark) {
+  .apexcharts-legend-text {
+    @apply !tce-text-slate-50
```
jantoun-scottlogic, line 7: Why do we need '! important' here?

mgriffin-scottlogic: I think because the class is used so far down the hierarchy and the chart has some method of setting the colour already, we need to override it. Doesn't work without it and any other methods get into conditionally setting things on the chart options in javascript etc.

jantoun-scottlogic: Ok, maybe !important is the most maintainable solution then.

jantoun-scottlogic (approved):

PR #92, opened 2024-06-24
SFD-127: Add tooltip to calculate button when its disabled

jantoun-scottlogic:
[SFD-127 Jira ticket](https://scottlogic.atlassian.net/browse/SFD-127)

Adds a message under the calculate button when its disabled to explain which fields are invalid

With one error:
one-error

Multiple errors:
multiple-errors
I think getting the text all the way to the right as in the one error case when the text wraps with multiple errors might require some JavaScript so I've left it like this for now. (Commas and full stops for the list of errors also omitted for now since I think that may also require some JavaScript including adding some logic to check which fields are invalid to the component.) Maybe just having it always be left aligned would be better.

At the moment its set up to always display the message when the form is invalid (as long as it's been interacted with) rather than being like a tooltip that is only visible when the calculate button is focussed. I thought that might work a bit better if we include links to the invalid fields but we could try other approaches.


src/app/carbon-estimator-form/carbon-estimator-form.component.html
```diff
@@ -9,11 +9,12 @@
         </label>
         <input
           id="headCount"
+          #headCountInput
           class="tce-border tce-border-slate-400 tce-rounded"
           formControlName="headCount"
           name="headCount"
           type="number"
-          min="1"
+          required
```
jantoun-scottlogic, line 17: I moved the min validation logic to component.ts so it wasn't split between component.html and component.ts. The [Angular docs](recommend) adding the required field to the html for accessibility reasons  even if it isn't required for validation because Validators.required is already being used.

src/app/carbon-estimator-form/carbon-estimator-form.component.html
```diff
@@ -80,7 +81,14 @@
...
           type="number"
-          min="0" />
+          required
+          aria-describedby="numberOfServersError" />
+        @if (numberOfServers?.invalid && (numberOfServers?.dirty || numberOfServers?.touched)) {
+          <div class="tce-text-error-red tce-flex tce-gap-1" aria-live="polite" id="numberOfServersError">
+            <span class="material-icons-outlined">error</span>
+            <p>The number of servers must be greater than or equal to 0.</p>
+          </div>
+        }
```
jantoun-scottlogic, line 91: Not part of the ticket but I thought it would be good to add an inline error message for the number of servers field like other similar fields.

mgriffin-scottlogic: I think this is fine, just worth noting that the aria-describedby will probably trigger another warning in WAVE, which we should handle in https://scottlogic.atlassian.net/browse/SFD-129

I wonder also if it's worth wrapping up the duplication of 3 checks on invalid/dirty/touched or if that's just the best way to deal with it. I'm not sure whether an 'InvalidPipe' would recognise internal changes to an object passed to it, I forget whether it does a deep comparison.

mgriffin-scottlogic (approved):

src/app/carbon-estimator-form/carbon-estimator-form.component.html
jantoun-scottlogic: I set the aria-describedby to only be set when the message is present which is preventing it triggering an error in WAVE. I couldn't get the more concise `"{ value : condition}"` logic to work unfortunately, maybe I'm just making a slight syntax error.

I had a go at making a pipe that checks to see if a control is invalid and it seemed to perform the check ok but I couldn't get it to work in an `@if` condition. I'm not sure if template expressions can be used there, I did a quick search and couldn't find an example of how to do it.

jantoun-scottlogic: I tried turning off 'format on save' but still couldn't get the `"{ value : condition}"` syntax to work.

jantoun-scottlogic: For checking whether a form control is invalid, we might be able to use a more general function rather than a pipe.

mgriffin-scottlogic: Weird, the pipe syntax should work in an `@if` as well - `@if (monthlyActiveUsers | invalid) {`.
However trying it out myself, what I was thinking was right - a pure pipe doesn't get recalled when the contents of the object changes, so you'd need to make it impure and then I think a general function is equivalent to it. Would definitely be nice to roll some of that up if it's used as a check in two places close together.

jantoun-scottlogic: Ah ok, yes that was the problem. For some reason I thought it was triggering when I put it in a template expression somewhere else and the issue was related to the `@if` but it wasn't.

mgriffin-scottlogic (approved):

PR #93, opened 2024-06-24
SFD-115: Add reports for automated tests

jantoun-scottlogic:
[SFD-115 Jira ticket](https://scottlogic.atlassian.net/browse/SFD-115)

Adds artifact uploads in the playwright workflow for the traces and an HTML report file of the test results.

Notes:
WRT the unit tests, whilst Web Test Runner does support reporting, it seems that the experimental Angular builder which is the bridge between Angular and Web Test Runner doesn’t yet support this feature and there isn’t any documentation it so we may have to wait a bit to add reports for the unit tests.

On the Playwright front, I’ve added artifact uploads for the Playwright traces for failing tests and an HTML report using the pytest-html package which can be downloaded. @qgogay-scottlogic and @sdun-scottlogic are going to have a discussion about what sort of things we’d like to do with the test reporting (e.g. do we want to try to add a display to the GitHub UI?) and we can then investigate further in a follow up ticket.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,6 +4,8 @@ colorama==0.4.6
 greenlet==3.0.3
 idna==3.7
 iniconfig==2.0.0
+Jinja2==3.1.4
+MarkupSafe==2.1.5
 packaging==24.1
 playwright==1.44.0
 pluggy==1.5.0
@@ -11,6 +13,8 @@ psutil==5.9.8
 pyee==11.1.0
 pytest==8.2.2
 pytest-base-url==2.1.0
+pytest-html==4.1.1
+pytest-metadata==3.1.1
 pytest-playwright==0.5.0
 pytest-xprocess==1.0.2
 python-slugify==8.0.4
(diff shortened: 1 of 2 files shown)


requirements.txt
```diff
@@ -4,13 +4,17 @@ colorama==0.4.6
...
+Jinja2==3.1.4
+MarkupSafe==2.1.5
 packaging==24.1
 playwright==1.44.0
 pluggy==1.5.0
 psutil==5.9.8
 pyee==11.1.0
 pytest==8.2.2
 pytest-base-url==2.1.0
+pytest-html==4.1.1
```
jantoun-scottlogic, line 16: [pytest-html](https://pypi.org/project/pytest-html/) is the main dependency. The other 3 packages added are transient dependencies.

mgriffin-scottlogic: As discussed yesterday, I think this is fine given Python is only needed for tests - we probably don't need to go for something like Poetry at this point.

mgriffin-scottlogic (approved):

PR #96, opened 2024-06-26
SFD-123: Add assumptions view option before calculation

jantoun-scottlogic:
[SFD-123 Jira ticket](https://scottlogic.atlassian.net/browse/SFD-123)

Moves the button to open the assumptions and limitations view to the estimator form below the disclaimer section.

new-button-position

Notes:
- When closing the assumptions and limitations view, the open button is scrolled into view expect when pressing escape when the view does not have focus in which case the old behaviour was retained.

### DIFF
--- a/playwright/tests/test_17_EstTool_UI_AssumpLimitations.py
+++ b/playwright/tests/test_17_EstTool_UI_AssumpLimitations.py
@@ -62,8 +62,8 @@ def test_example(page: Page) -> None:
     expect(page.get_by_role("button", name="Assumptions and limitations")).to_be_visible()
     page.get_by_role("button", name="Assumptions and limitations").click()
     expect(page.get_by_role("heading", name="Assumptions and Limitations")).to_be_visible()
-    expect(page.get_by_label("Close assumptions and").filter(has_not_text=re.compile("Close"))).to_be_visible() # 'X' close button
-    # Need regex because has_text="Close" is case insensitive and 'X' button has the text 'close'
+    expect(page.get_by_label("Close assumptions and").filter(has_text=re.compile("close"))).to_be_visible() # 'X' close button
+    # Need regex because has_text="close" is case insensitive and other close button has the text 'Close'
     expect(page.get_by_text("The Technology Carbon Estimator tool is designed to")).to_be_visible()
     expect(page.get_by_role("heading", name="Assumptions", exact=True)).to_be_visible()
     expect(page.get_by_role("heading", name="Time period")).to_be_visible()
--- a/playwright/tests/test_17_EstTool_UI_AssumpLimitations.py
+++ b/playwright/tests/test_17_EstTool_UI_AssumpLimitations.py
@@ -62,7 +62,9 @@ def test_example(page: Page) -> None:
     expect(page.get_by_role("button", name="Assumptions and limitations")).to_be_visible()
     page.get_by_role("button", name="Assumptions and limitations").click()
     expect(page.get_by_role("heading", name="Assumptions and Limitations")).to_be_visible()
-    expect(page.get_by_text("The Technology Carbon")).to_be_visible()
+    expect(page.get_by_label("Close assumptions and").filter(has_not_text=re.compile("Close"))).to_be_visible() # 'X' close button
+    # Need regex because has_text="Close" is case insensitive and 'X' button has the text 'close'
+    expect(page.get_by_text("The Technology Carbon Estimator tool is designed to")).to_be_visible()
     expect(page.get_by_role("heading", name="Assumptions", exact=True)).to_be_visible()
     expect(page.get_by_role("heading", name="Time period")).to_be_visible()
     expect(page.get_by_text("The estimation is based on a")).to_be_visible()
@@ -96,7 +98,6 @@ def test_example(page: Page) -> None:
     expect(page.get_by_text("These figures are combined")).to_be_visible()
     expect(page.get_by_role("heading", name="Network Data Transfer")).to_be_visible()
     expect(page.get_by_text("Our outgoing network data")).to_be_visible()
-    expect(page.get_by_text("Assumptions and Limitations The Technology Carbon Estimator tool is designed to")).to_be_visible()
     page.get_by_role("heading", name="End-User Devices").click()
     expect(page.get_by_text("We combine device information")).to_be_visible()
     expect(page.get_by_role("heading", name="Limitations", exact=True)).to_be_visible()
@@ -113,8 +114,8 @@ def test_example(page: Page) -> None:
     expect(page.get_by_text("Like Off the shelf and open")).to_be_visible()
     expect(page.get_by_role("heading", name="Managed Services")).to_be_visible()
     expect(page.get_by_text("We currently do not make a")).to_be_visible()
-    expect(page.get_by_label("Close assumptions and")).to_be_visible()
-    page.get_by_label("Close assumptions and").click()
+    expect(page.get_by_label("Close assumptions and").filter(has_text=re.compile("Close"))).to_be_visible()
+    page.get_by_label("Close assumptions and").filter(has_text=re.compile("Close")).click()
     expect(page.get_by_role("heading", name="Technology Carbon Estimator")).to_be_visible()
     expect(page.get_by_role("heading", name="Organisation")).to_be_visible()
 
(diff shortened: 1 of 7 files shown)


playwright/tests/test_17_EstTool_UI_AssumpLimitations.py
```diff
@@ -96,7 +96,8 @@ def test_example(page: Page) -> None:
     expect(page.get_by_text("These figures are combined")).to_be_visible()
     expect(page.get_by_role("heading", name="Network Data Transfer")).to_be_visible()
     expect(page.get_by_text("Our outgoing network data")).to_be_visible()
-    expect(page.get_by_text("Assumptions and Limitations The Technology Carbon Estimator tool is designed to")).to_be_visible()
+    expect(page.get_by_text("Assumptions and Limitations")).to_be_visible()
```
jantoun-scottlogic, line 99: I needed to break this up because it was failing. I think it was because there is now an 'X' button between the first and second sentences.

playwright/tests/test_17_EstTool_UI_AssumpLimitations.py
```diff
@@ -96,7 +98,6 @@ def test_example(page: Page) -> None:
     expect(page.get_by_text("These figures are combined")).to_be_visible()
     expect(page.get_by_role("heading", name="Network Data Transfer")).to_be_visible()
     expect(page.get_by_text("Our outgoing network data")).to_be_visible()
-    expect(page.get_by_text("Assumptions and Limitations The Technology Carbon Estimator tool is designed to")).to_be_visible()
```
jantoun-scottlogic, line 99: This was causing the test to fail, I think because there is now an 'X' button in between the first and second halves of the test. I removed this expectation because I think it's a duplicate of some of the expectations [above](https://github.com/ScottLogic/sl-tech-carbon-estimator/pull/96/files#diff-0574963fc62f5929ba3c969da81828750d7d54fbcf5b883e54416d6c983313e2R64).

src/app/assumptions-and-limitation/assumptions-and-limitation.component.spec.ts
```diff
@@ -41,26 +40,14 @@ describe('AssumptionsAndLimitationComponent', () => {
...
 
     component.ngAfterContentInit();
     fixture.detectChanges();
 
     document.dispatchEvent(new KeyboardEvent('keydown', { key: 'Escape' }));
 
-    expect(component.closeEvent.emit).toHaveBeenCalledWith(true);
-  });
-
-  it('should emit close event with false value when press esc key with focus not within component', () => {
```
jantoun-scottlogic, line 55: This test was testing the opposite scenario to the one in the description. I'm not sure how to drop focus on the component without testing from a parent component.

src/app/tech-carbon-estimator/tech-carbon-estimator.component.spec.ts
```diff
@@ -103,42 +103,15 @@ describe('TechCarbonEstimatorComponent', () => {
...
-    component.showAssumptionsAndLimitationView = true;
-    fixture.detectChanges();
-    spyOn(component.assumptionsLimitation.nativeElement, 'scrollIntoView').and.callThrough();
-    component.showAssumptionsAndLimitation();
-    fixture.detectChanges();
-
-    expect(component.assumptionsLimitation.nativeElement.scrollIntoView).toHaveBeenCalled();
-  });
-
-  it('should scroll to top of estimations when closeAssumptionsAndLimitation is called', () => {
```
jantoun-scottlogic, line 117: The equivalent test would now be to check that the open button is scrolled into view when this is called but the open button is only on the page when the assumptions and limitations view is closed it can't be spied on before the close button is pressed.

src/app/tech-carbon-estimator/tech-carbon-estimator.component.html
```diff
@@ -9,6 +9,14 @@ <h1 class="tce-text-3xl tce-mb-6">Technology Carbon Estimator</h1>
           aria-live="polite"
           (closeEvent)="closeAssumptionsAndLimitation($event)"></assumptions-and-limitation>
       } @else {
+        <div class="tce-flex tce-justify-end tce-pb-4 -tce-mt-2">
+          <button
+            #showAssumptionsLimitationButton
+            class="tce-button-assumptions tce-px-3 tce-w-fit tce-self-end"
```
mgriffin-scottlogic, line 15: Was it intentional to drop the `tce-py-2` here? I'm not sure if it had an effect on the Scott Logic styling but just looking at the npm-package styling it's now a bit small:

image

src/app/assumptions-and-limitation/assumptions-and-limitation.component.ts
```diff
@@ -82,16 +87,16 @@ export class AssumptionsAndLimitationComponent implements AfterContentInit {
   }
 
   public ngAfterContentInit(): void {
-    this.assumptionsLimitation.nativeElement.focus();
+    this.assumptionsLimitation.nativeElement.focus({ preventScroll: true });
   }
 
-  public onClose(hasFocus = true): void {
-    this.closeEvent.emit(hasFocus);
+  public onClose(focusOpenButton: boolean, scrollToOpenButton: boolean): void {
```
mgriffin-scottlogic, line 93: Would it be possible to have this function use the `AssumptionsLimitationCloseEvent` as its parameter? Fine if it's awkward or doesn't work in the html, it's just a bit confusing seeing (true, false) or (true, true) with no context.

src/app/tech-carbon-estimator/tech-carbon-estimator.component.spec.ts
mgriffin-scottlogic: I guess if we wanted to keep this test we could hide the button rather than make it non-existent. Or scroll to something that does always exist, though the behaviour of this is starting to feel a bit complicated and I wouldn't want to add more.
To be honest, I'm not sure if the scroll is ever needed, as you end up in the right place if the button is focussed. The only time it doesn't try to focus is if you hit escape while focussed outside of the assumptions - which means you've either clicked off it or tabbed all the way through to the diagram. Do we really care where the form is scrolled to in that case? Or it could be scroll is the alternative to focus rather than allowing both separately.

mgriffin-scottlogic (changes requested):

src/app/tech-carbon-estimator/tech-carbon-estimator.component.html
jantoun-scottlogic: If I remember correctly this was intentional and it was because the spacing looked a bit large with the Scott Logic styling. I'll have a look at both versions and see what I can do.

mgriffin-scottlogic: Hmm, I wonder if something's caused a priority change, as it looks like the sl-styles.css is already trying to override padding to a small amount for this class.

src/app/tech-carbon-estimator/tech-carbon-estimator.component.spec.ts
jantoun-scottlogic: Yes, it seems the explicit scrolling is not required. I've removed it so it just has the focus behaviour.

src/app/tech-carbon-estimator/tech-carbon-estimator.component.html
jantoun-scottlogic: I've added the padding back in. I think it actually looks fine with the Scott Logic styling.

mgriffin-scottlogic: Yep, looks good to me too. Could have been something odd with styles updating while still serving locally, doesn't always seem to update fully.

src/app/assumptions-and-limitation/assumptions-and-limitation.component.ts
jantoun-scottlogic: I've reverted the function signature to `onClose(hasFocus: boolean = true)` now that the event has been reverted to just consist of the `hasFocus` boolean.

src/app/assumptions-and-limitation/assumptions-and-limitation.component.spec.ts
mgriffin-scottlogic: Yeah, I think you'd have to set up a test component to contain it and go through it. You'd also be testing that nothing happens now essentially - if you hit escape with diagram focussed, diagram remains focussed. Could be that's easier to do in Playwright?

mgriffin-scottlogic (approved):

src/app/assumptions-and-limitation/assumptions-and-limitation.component.spec.ts
jantoun-scottlogic: Yes this one might be best left out of the unit tests. We could add a Playwright test if we would like to test this scenario.

PR #97, opened 2024-06-28
SFD-136 Set up workflow to publish npm package

mgriffin-scottlogic:
Final process should be:
- Run tests
- Build package and upload as artifact
- Use artifact to publish to npm
- Use artifact and version tag to create release

It's not enforced but the intention is that you run the workflow manually using the corresponding tag/commit for the version.

Also updated the pages workflow so that it is triggered when a new release is created.

Tested as far as possible to ensure that npm publish is authenticated and draft release is created but need to run from main with an updated version for it to complete fully.

### DIFF
--- a/.github/workflows/publish-package.yml
+++ b/.github/workflows/publish-package.yml
@@ -1,6 +1,7 @@
 name: Publish new package
 on:
   workflow_dispatch:
+  pull_request:
 env:
   PACKAGE_PREFIX: scottlogic-tech-carbon-estimator-
 concurrency:
@@ -61,7 +62,7 @@ jobs:
         env:
           NODE_AUTH_TOKEN: ${{secrets.NPM_TOKEN}}
         run: |
-          npm publish ${PACKAGE_PREFIX}${VERSION}.tgz --provenance --access public
+          npm publish ${PACKAGE_PREFIX}${VERSION}.tgz --provenance --access public --dry-run
 
   release:
     runs-on: ubuntu-latest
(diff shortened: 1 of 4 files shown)


.github/workflows/publish-package.yml
```diff
@@ -0,0 +1,88 @@
...
+          npm pkg delete scripts private devDependencies files
+          npm pkg set main=main.js
+          npm pack
+          VERSION=$(npm pkg get version --workspaces=false | tr -d \")
+          echo "VERSION=$VERSION" >> $GITHUB_ENV
+      - name: Upload artifact
+        uses: actions/upload-artifact@v4
+        with:
+          if-no-files-found: error
+          name: ${{env.PACKAGE_PREFIX}}${{env.VERSION}}
```
mgriffin-scottlogic, line 37: I think this name could actually just be 'package' rather than including the full name/version, given the artefacts are specific to the workflow run. Only reduces the use of these env vars by a couple of places though and I don't know if it's nicer to have it named for its contents if you go looking at the workflow in the UI?

jantoun-scottlogic: I think either would be fine. I quite like seeing the name and version in the UI.

jantoun-scottlogic: Maybe a 'package_name' variable could help a bit with variable usage if that's easy to set up but I think it's fine like it is.

mgriffin-scottlogic: Yeah, I think it's reasonably clear like this. If I set up another var it has to be added to this job, made an output of it, and included in the env section of each following job.

jantoun-scottlogic (approved):

sdun-scottlogic:
Tested & Approved - SD.

PR #98, opened 2024-07-01
SFD-139: Add an inline error message for the' number of servers' field

jantoun-scottlogic:
[SFD-139 Jira ticket](https://scottlogic.atlassian.net/browse/SFD-139)

Adds an inline error message for the 'number of servers' field. Also adds a pipe to check whether a form control (or group) is invalid and the user has interacted with it.

number-of-servers-error


src/app/carbon-estimator-form/carbon-estimator-form.component.spec.ts
```diff
@@ -50,14 +50,21 @@ describe('CarbonEstimatorFormComponent', () => {
   describe('headCount()', () => {
     it('should not return null once the component is initialized', () => {
       component.ngOnInit();
-      expect(component.headCount).not.toBe(null);
+      expect(component.headCount === null).toBeFalse();
```
mgriffin-scottlogic, line 53: Any particular reason to change these to `=== null`? I would have thought the previous method would give a more understandable error - `Expected null not to be null.` vs  `Expected true to be false.`

Possible they're both not that helpful, could add
`expect(component.headCount).withContext('headCount').not.toBe(null);`
which outputs
`headCount: Expected null not to be null.`

jantoun-scottlogic: The issue was that it wouldn't compile using `.not.toBe(null)` because I've coerced the return values of the methods to not include `null` by adding a `!` after the `.get` because they should never return `null` once the form group has been created in `ngOnInit`. This was to avoid needing null checks when using the getters/to not have to pass a nullable argument to the pipe. Using `=== null` was a workaround for the tests. I'm slightly surprised it doesn't complain about the fact that the return type doesn't include `null` here as well. I'll see if I can improve the error message and add a comment if not.

mgriffin-scottlogic: That's odd, I thought it might be around the coerced null but it didn't seem to have a problem with it for me.

jantoun-scottlogic: There may also be an option to suppress the compiler error on these lines and use `not.toBe(null)`.

jantoun-scottlogic: It seems fine with `.not.toBeNull()` so I've changed it to that. There is also [@ts-expect-error](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-9.html#-ts-expect-error-comments) which would have suppressed the error for `.not.toBe(null)`.

Unfortunately, the pipe throws an error when a getter is broken which causes all the tests for the component to fail without a clear message. I'm not sure what to do about that.

jantoun-scottlogic: It's also causing the unit test suite to hang when they fail.

mgriffin-scottlogic: Ah, well maybe you can't expect to be able to deal with everything - to reach that point I assume you had to deliberately return null while saying that it could never be null?

I guess the options are to allow the pipe to take null and immediately return false or try to mock out the pipe for the unit test but that seems a bit extreme.

src/app/pipes/invalid-and-interaction.pipe.spec.ts
```diff
@@ -0,0 +1,45 @@
...
+
+  it('should return false if the control is invalid and there has been no interaction', () => {
+    const formControl = new FormControl('', Validators.required);
+    const pipe = new InvalidAndInteractionPipe();
+
+    const transformed = pipe.transform(formControl);
+    expect(transformed).toBeFalse();
+  });
+
+  it('should return true if the form is invalid and there has been interaction', () => {
```
mgriffin-scottlogic, line 36: Minor point but to cover everything, the dirty and touched states should be tested in isolation.
Also don't know if it's worth testing against some stub/mock AbstractControl instead of a concrete FormControl, so you're not relying on its behaviour.

mgriffin-scottlogic (approved):

src/app/pipes/invalid-and-interaction.pipe.spec.ts
jantoun-scottlogic: I've separated those cases and replaced the form controls with (partial) jasmine mocks of abstract controls. I added a few complementary properties to the ones used in the implementation of the pipe to try to cover a few possible implementation changes. I also parameterised the tests.

mgriffin-scottlogic (approved):

src/app/carbon-estimator-form/carbon-estimator-form.component.html
```diff
@@ -12,8 +12,8 @@
           name="headCount"
           type="number"
           min="1"
-          aria-describedby="headCountError" />
-        @if (headCount?.invalid && (headCount?.dirty || headCount?.touched)) {
+          [attr.aria-describedby]="(headCount | invalidAndInteraction) ? 'headCountError' : null" />
```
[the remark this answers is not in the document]
jantoun-scottlogic: I've renamed the pipe to `invalidated` and added a null check to resolve some of the testing and component initialisation issues if a null value does creep through somehow as discussed.

mgriffin-scottlogic (approved):

PR #100, opened 2024-07-02
SFD-136 Publish pages after successful release

mgriffin-scottlogic:
- Allow pages workflow to be called from other workflows
- Restrict publish workflow so that it must be triggered from a matching tag
- Tidy up package name environment variables

### DIFF
--- a/.github/workflows/publish-package.yml
+++ b/.github/workflows/publish-package.yml
@@ -86,3 +86,6 @@ jobs:
               --repo="$GITHUB_REPOSITORY" \
               --prerelease \
               --generate-notes
+
+  publish_pages:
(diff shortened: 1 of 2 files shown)


.github/workflows/publish-package.yml
```diff
@@ -1,20 +1,50 @@
 name: Publish new package
 on:
   workflow_dispatch:
+  pull_request:
```
mgriffin-scottlogic, line 4: Will remove this before merging

mgriffin-scottlogic:
Appreciate these are kind of hard to review but you can look at the last failed job to see how it should prevent incorrect usage. Should be able to test fully with next release.

PR #104, opened 2024-07-05
SFD-113: Display a placeholder for the diagram before it is shown

jantoun-scottlogic:
[SFD-113 Jira ticket](https://scottlogic.atlassian.net/browse/SFD-113)

Adds a placeholder chart where the estimation chart will be shown when the form is submitted. The placeholder is displayed initially and when the form is reset.

Placeholder:

placeholder-chart

Estimation:

estimation-chart


src/app/carbon-estimation/carbon-estimation.component.ts
```diff
@@ -48,12 +43,20 @@ export class CarbonEstimationComponent implements OnInit, OnDestroy {
 
   constructor(private changeDetectorRef: ChangeDetectorRef) {
     effect(() => {
-      this.emissions = this.getOverallEmissionPercentages(this.carbonEstimation());
-      this.emissionAriaLabel = this.getAriaLabel(this.emissions);
+      if (this.showPlaceholderDiagram) {
```
mgriffin-scottlogic: I wonder if it would be better to just use the carbonEstimation input here - store it locally before the if. That way Typescript should know that it can't be null when you set it in the chartData with the ! added.

src/app/carbon-estimation/carbon-estimation.component.ts
```diff
@@ -48,12 +43,20 @@ export class CarbonEstimationComponent implements OnInit, OnDestroy {
...
+      } else {
+        this.chartData = this.getOverallEmissionPercentages(this.carbonEstimation()!);
+        this.chartOptions = getChartOptions(false);
+        this.emissionAriaLabel = this.getAriaLabel(this.chartData);
+      }
     });
   }
 
   public ngOnInit(): void {
+    this.chartOptions = getChartOptions(this.carbonEstimation() === null);
```
mgriffin-scottlogic, line 59: Should this section be part of the effect above? Given it sets the chartOptions to a fresh object whenever the input carbonEstimation changes.

src/app/carbon-estimation/carbon-estimation.component.ts
jantoun-scottlogic: Yes, that seems neater. I've made that change.

src/app/carbon-estimation/carbon-estimation.component.ts
jantoun-scottlogic: I think this line should perhaps go in the constructor (and the argument to `getChartOptions` needs to be updated). I think it could go either there on in `ngOnInit` but perhaps the constructor is a more natural place to initialise it and it let's you drop the `!` in the variable declaration.

I'm not sure why but it seemed like setting the chart height when the chartOptions were overwritten in the effect wasn't required. It seemed to work fine when resizing and swapping between the estimation and placeholder. I'll look into it a bit more and see if I can work out what's going on. Perhaps it's still worth setting the chart height in the effect just in case.

jantoun-scottlogic: Nvm, it can't go in the constructor because the input is not available there :^) (It just looked like it worked because the initial value passed is currently `undefined`.)

jantoun-scottlogic: I've moved the chart configuration logic to a separate method which is called in `ngOnInit` and when the input is changed.

src/app/carbon-estimation/carbon-estimation.component.css
```diff
@@ -16,25 +16,29 @@
 }
 
 ::ng-deep .web-logo {
-  background-image: url("data:image/svg+xml,%3Csvg xmlns='http://www.w3.org/2000/svg' viewBox='0 -960 960 960'%3E%3Cpath style='fill:white' d='M160-160q-33 0-56.5-23.5T80-240v-480q0-33 23.5-56.5T160-800h640q33 0 56.5 23.5T880-720v480q0 33-23.5 56.5T800-160H160Zm0-80h420v-140H160v140Zm500 0h140v-360H660v360ZM160-460h420v-140H160v140Z'/%3E%3C/svg%3E");
+  background-image: url('../../assets/web.svg');
```
jantoun-scottlogic, line 19: I moved the SVGs to separate files because Angular was throwing an error due to the size of this CSS file when building the app. (There is also the option to increase the allowed file size in the angular.json file.)

src/app/carbon-estimation/carbon-estimation.constants.ts
```diff
@@ -7,80 +7,102 @@ export enum EmissionsColours {
   Downstream = '#4B7E56',
 }
 
+export enum PlaceholderEmissionsColours {
+  Upstream = '#333',
+  Direct = '#666',
+  Indirect = '#888',
+  Downstream = '#AAA',
+}
```
jantoun-scottlogic, line 15: I haven't tried to match the greyscale for the placeholder chart to the shades in the actual chart.

src/app/carbon-estimation/carbon-estimation.component.css
mgriffin-scottlogic: Just had a quick check and this seems to work as expected. I think initially we were concerned that the constantly changing tooltip would make a request for the image each time. Also looks to include the images when creating the npm package.

src/app/carbon-estimation/carbon-estimation.constants.ts
mgriffin-scottlogic: I think that's fine, doing it from the real colours might not actually give enough difference between the colours to make them identifiable.

src/app/carbon-estimation/carbon-estimation.component.ts
mgriffin-scottlogic: I think we could still remove the call in `ngOnInit()` but we need to initialise the `chartOptions` on construction ie.

`public chartOptions: ChartOptions = getChartOptions(true);`

Without that, you end up with an error in the template that must prevent the signal/effect setup, meaning it never gets changes passed through.
```
TypeError: Cannot read properties of undefined (reading 'chart')
    at CarbonEstimationComponent_Template (carbon-estimation.component.html:28:27)
```
Bit of an optional one, you're either specifying the `chartOptions` in two places or calling `configureChart()` in two places.

src/app/tech-carbon-estimator/tech-carbon-estimator.component.spec.ts
```diff
@@ -90,17 +76,28 @@ describe('TechCarbonEstimatorComponent', () => {
...
     expect(estimationServiceStub.calculateCarbonEstimation).toHaveBeenCalledWith(formValue);
-    expect(component.showEstimation).toBeTrue();
   });
 
-  it('should hide estimation if form is reset', () => {
-    component.showEstimation = true;
+  it('should set the carbonEstimation when handleFormSubmit is called', () => {
+    component.carbonEstimation = null;
+    fixture.detectChanges();
+    spyOn(estimationServiceStub, 'calculateCarbonEstimation' as never).and.callThrough();
```
mgriffin-scottlogic, line 86: Had to go and look up what `as never` was for. There's a more explicit alternative, which I don't know whether it's any clearer?

`spyOn<Partial<CarbonEstimationService>, any>(estimationServiceStub, 'calculateCarbonEstimation').and.callThrough();`

mgriffin-scottlogic (dismissed):

src/app/carbon-estimation/carbon-estimation.component.ts
jantoun-scottlogic: Are you seeing the error in the current setup or is that something that would need to be handled with the proposed change?

I think we probably need some chart configuration in `ngOnInit` because it depends on the `extraHeight` and `carbonEstimation` signals and I don't think they're available during the initial construction. I think it's going to be easier to do all of the initial setup in one place rather than split it between construction and `ngOnInit` and we can't configure the height without the `extraHeight` signal (it's used in `getChartHeight` which is used in `configureChart`)*.

Also, if we make a choice to set the chart options to the placeholder options on construction rather than basing it off the input we'd have a setup where sometimes the options are based on the input and sometimes they are a based on a choice we hard coded into the component and I think that's going to be potentially confusing when using the component. I think they should always depend on the input.

 *I think one thing we should look into refactoring is having the extraHeight be an explicit numerical argument that is passed into functions where it's used. At the moment the signal is accessed directly inside the `getChartHeight` function so it's not obvious when it's called that this function depends on a signal. I think it should also be a numerical input to the component rather than a string that is transformed inside the carbon-estimation component.

jantoun-scottlogic: Having said that, I think it would be nice to make it more explicit that `configureChart` is setting the `chartOptions`, it's not obvious that an assignment to the `chartOptions` variable is taking place when calling `configureChart`.

src/app/tech-carbon-estimator/tech-carbon-estimator.component.spec.ts
jantoun-scottlogic: Ah, good spot, I think that line is actually unnecessary. I think it should be sufficient to mock the injected dependency via `TestBed.configureTestingModule` and it's probably best to avoid spying on private methods when possible. I've deleted it.

src/app/carbon-estimation/carbon-estimation.component.ts
mgriffin-scottlogic: Sorry for the confusion, I only see the error if I remove `this.configureChart(this.carbonEstimation());` from `ngOnInit()`, without also setting the `chartOptions` to something valid.

The effect still seems to get called (probably just after `ngOnInit()`) - so even if I initialise `chartOptions` to the non-placeholder options, it will update to use the grey one and you never actually see the colours. May be behaviour that we shouldn't rely on tho.

mgriffin-scottlogic: Also if we want to use signals properly this should maybe use `computed()` instead of `effect()` but that's a bigger change.

jantoun-scottlogic: Ah, for some reason I thought the effect wasn't running when the component is initialised but it does indeed seem to be running after `ngOnInit`. Even though it looks like the initial setup will be immediately overwritten by the effect running I think it will still be clearer to do the initialisation in `ngOnInit` based on the inputs. As you alluded to, I think that will also be more robust if the behaviour of `effect()` changed in the future or we swapped to `computed()` and that didn't run after initialisation.

mgriffin-scottlogic (approved):

src/app/carbon-estimation/carbon-estimation.component.spec.ts
```diff
@@ -47,16 +47,11 @@ describe('CarbonEstimationComponent', () => {
     fixture.detectChanges();
   });
 
-  it('should set the chart height when the component is initialised', () => {
-    spyOnProperty(window, 'innerHeight').and.returnValue(1000);
-    spyOnProperty(window, 'innerWidth').and.returnValue(1000);
-    spyOnProperty(window.screen, 'height').and.returnValue(1080);
-    spyOnProperty(component.detailsPanel.nativeElement, 'clientHeight').and.returnValue(200);
```
jantoun-scottlogic, line 54: Now that the height isn't set in `ngOnInit` these spies would need to be set in the `beforeEach` section to take effect in time. Whilst it works for this test I think it might seem a bit odd to explicitly set the dimensions to desktop dimensions before every test since some are describing mobile behaviour. I've changed the test to just check that the height is set to a positive value initially.

mgriffin-scottlogic: Yeah, seems reasonable.

sdun-scottlogic:
Tested & Approved - SD.

PR #107, opened 2024-07-10
SFD-118 Improve accessibility of assumptions footnote

mgriffin-scottlogic:
Add link to/from network embodied carbon note, with appropriate aria roles.

### DIFF
--- a/src/app/assumptions-and-limitation/assumptions-and-limitation.component.html
+++ b/src/app/assumptions-and-limitation/assumptions-and-limitation.component.html
@@ -111,7 +111,7 @@
   <h4 class="tce-text-lg">Upstream Emissions</h4>
   <p class="tce-text-sm">
     Each class of device is given an average amount of embodied carbon and a lifespan, based on averages taken from
-    manufacturer provided Product Carbon Footprint documentation<a id="networkNoteRef" role="doc-noteref" href="#networkNote">*</a>. At present these are:
+    manufacturer provided Product Carbon Footprint documentation<a id="networkNoteRef" role="doc-noteref" href="#networkNote" aria-label="see footnote">*</a>. At present these are:
   </p>
   <table class="tce-w-fit">
     <thead>
--- a/src/app/assumptions-and-limitation/assumptions-and-limitation.component.html
+++ b/src/app/assumptions-and-limitation/assumptions-and-limitation.component.html
@@ -138,7 +138,7 @@
     </tbody>
   </table>
   <aside id="networkNote" class="tce-text-sm tce-italic" role="doc-footnote">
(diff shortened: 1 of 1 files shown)


src/app/assumptions-and-limitation/assumptions-and-limitation.component.html
```diff
@@ -111,7 +111,7 @@ <h4 class="tce-text-lg">Carbon Intensity</h4>
   <h4 class="tce-text-lg">Upstream Emissions</h4>
   <p class="tce-text-sm">
     Each class of device is given an average amount of embodied carbon and a lifespan, based on averages taken from
-    manufacturer provided Product Carbon Footprint documentation*. At present these are:
+    manufacturer provided Product Carbon Footprint documentation<a id="networkNoteRef" role="doc-noteref" href="#networkNote">*</a>. At present these are:
```
jantoun-scottlogic, line 114: It it worth adding something like `aria-label="footnote"`?

mgriffin-scottlogic: Yeah, I don't see a problem with that. NVDA and Narrator both seem to vary on whether it announces the note reference role, at least this would be explicit. Maybe it should be `"see footnote"` as the footnote itself will also say 'footnote'?

jantoun-scottlogic: Ah ok, I only checked with NVDA. Yes, maybe "see footnote" would be clearer. Perhaps a label on the other `*` clarifying that it's a backlink would be good as well. What do you think?

mgriffin-scottlogic: I'd be inclined to leave that one as it is? Again it varies, but if I add a label there then Narrator will say 'backlink backlink asterisk' as it's already taking the role into account. Maybe put it through testing and see if any more suggestions come up?

jantoun-scottlogic: I wonder if my Narrator setting are different or I'm navigating the page in a different way to you. It says 'link asterisk' when I tab to the backlink rather than 'backlink...'. I agree that we don't want to have the words 'link' or 'backlink' explicitly in the label because they will normally be announced by the screen reader. I was thinking of trying to make it announce something like 'link/backlink return to main text' rather than 'link asterisk'.

mgriffin-scottlogic: It does seem to behave differently depending on how you navigate the page, I couldn't seem to get the response I was seeing yesterday again. I might add a label that will make sense as you tab through it at least.

jantoun-scottlogic (approved):

sdun-scottlogic:
Tested & Approved - SD.

PR #108, opened 2024-07-10
SFD-135 Set up workflow to run when merging PR's

jmain-scottlogic:
[SFD-135](https://scottlogic.atlassian.net/browse/SFD-135)

- Change pull-request workflow to build and test.
- Update so runs when PR's merged into main
- Removed playwright workflow as covered by build and test
- Update test step in pages to run build and test workflow


mgriffin-scottlogic (changes requested):
We'll also need to change [publish-package.yml](https://github.com/ScottLogic/sl-tech-carbon-estimator/blob/1d06a7c8385df5f3b8ea84308182bb82948e643a/.github/workflows/publish-package.yml#L38) as this calls the old pull-request.yml workflow as well. Not sure exactly what happens when a workflow is called twice as this calls the page publish too, will have to see if it does something sensible or builds and tests things twice.

jmain-scottlogic:
Updated publish to point at build-and-test. Quickly looking into making the build-and-test step optional, but hit problem as github workflows don't seem to support conditional needs

mgriffin-scottlogic (approved):
Looks good, yeah Github Actions never seems to make it easy. There's some things you can do with conditionals but a lot of stuff is very easy to break and not know about it until things just don't run.

PR #109, opened 2024-07-12
SFD-145 Increase available space for SL hosted estimator

jmain-scottlogic:
Remove padding from within tech-carbon-estimator component. Added it back via the index file.

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -13,7 +13,7 @@
     <noscript
       >Javascript appears to be disabled - unfortunately the estimator requires it to be enabled to function.</noscript
     >
-    <div class="tce-px-4 tce-pt-4 pd:tce-pb-4">
+    <div class="tce-pt-4 md:tce-pb-4">
       <tech-carbon-estimator></tech-carbon-estimator>
     </div>
   </body>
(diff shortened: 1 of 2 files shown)


src/index.html
```diff
@@ -13,6 +13,8 @@
     <noscript
       >Javascript appears to be disabled - unfortunately the estimator requires it to be enabled to function.</noscript
     >
-    <tech-carbon-estimator></tech-carbon-estimator>
+    <div class="tce-px-4 tce-pt-4 pd:tce-pb-4">
```
mgriffin-scottlogic, line 16:
```suggestion
    <div class="tce-pt-4 md:tce-pb-4">
```
Assume typo on the bottom padding and I'm thinking we actually don't want the x padding here as the SL site is using this version within the iframe. We'll just have to be aware that it will look very close to the sides when testing in isolation.

jmain-scottlogic: It hadnt quite clicked that the SL site was using this as the iframe. I dont wonder if we need any padding then?

mgriffin-scottlogic (approved):

PR #110, opened 2024-07-12
SFD-122 Fix Axe accessibility issues

mgriffin-scottlogic:
- Add `<main>` landmarks to index pages - not strictly a problem as it depends where component used but makes it easier to test.
- Use `<figure>` element instead of a `<div>` as `role="presentation"` does not officially support labels.
- Remove `role="region"` from expansion panel as they should have a unique label.

Last one is the only possible controversial change, I'm not sure if there was a reason it originally had a region role?

### DIFF
--- a/src/app/expansion-panel/expansion-panel.component.html
+++ b/src/app/expansion-panel/expansion-panel.component.html
@@ -9,6 +9,6 @@
     {{ expanded ? 'expand_less' : 'expand_more' }}
   </button>
 </div>
-<div [ngClass]="{ 'tce-hidden': !expanded }" [class]="config.contentContainerStyles" role="region">
+<div [ngClass]="{ 'tce-hidden': !expanded }" [class]="config.contentContainerStyles">
   <ng-content select="[content]"></ng-content>
 </div>
--- a/src/package-index.html
+++ b/src/package-index.html
@@ -8,10 +8,12 @@
     <link href="https://fonts.googleapis.com/icon?family=Material+Icons+Outlined" rel="stylesheet" />
   </head>
   <body class="tce-m-0 dark:tce-bg-slate-800 dark:tce-text-slate-50">
-    <noscript
-      >Javascript appears to be disabled - unfortunately the estimator requires it to be enabled to function.</noscript
-    >
-    <tech-carbon-estimator></tech-carbon-estimator>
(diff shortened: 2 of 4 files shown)


src/app/expansion-panel/expansion-panel.component.html
```diff
@@ -9,6 +9,6 @@
     {{ expanded ? 'expand_less' : 'expand_more' }}
   </button>
 </div>
-<div [ngClass]="{ 'tce-hidden': !expanded }" [class]="config.contentContainerStyles" role="region">
+<div [ngClass]="{ 'tce-hidden': !expanded }" [class]="config.contentContainerStyles">
```
jantoun-scottlogic, line 12: I'm not sure why there was a region role here but I get a warning from SonarLint saying 'Use <section> instead of the region role to ensure accessibility across all devices.' so perhaps we shouldn't be using `role="region"` here regardless of whether a role is required. I would guess it would be down to the content that is passed in to have the correct roles so a `div` is ok here but I'm not sure.

jantoun-scottlogic (dismissed):

src/package-index.html
```diff
@@ -8,10 +8,12 @@
     <link href="https://fonts.googleapis.com/icon?family=Material+Icons+Outlined" rel="stylesheet" />
   </head>
   <body class="tce-m-0 dark:tce-bg-slate-800 dark:tce-text-slate-50">
-    <noscript
-      >Javascript appears to be disabled - unfortunately the estimator requires it to be enabled to function.</noscript
-    >
-    <tech-carbon-estimator></tech-carbon-estimator>
-    <footer class="tce-p-5 tce-bg-slate-800 tce-text-slate-50 dark:tce-bg-slate-950">© Copyright Scott Logic</footer>
+    <main>
```
jantoun-scottlogic, line 11: Should this `main` element have the same padding as `src/index.html`?

mgriffin-scottlogic: I think it's fine for this one to stay as it is? This one is just really for testing any issues that might come up when using it on the TCS site, without having to make a package and update it.

jantoun-scottlogic (dismissed):

sdun-scottlogic:
Tested & Approved - SD.

sdun-scottlogic:
Tested & Approved - SD.

jantoun-scottlogic (approved):

PR #113, opened 2024-07-18
SFD-143: Make dark mode error text easier to read

jantoun-scottlogic:
[SFD-143 Jira ticket](https://scottlogic.atlassian.net/browse/SFD-143)

Changes the the error messages in dark mode to white text on a red background so they are easier to read.

dark-mode-error-message

### DIFF
--- a/src/styles.css
+++ b/src/styles.css
@@ -3,11 +3,11 @@
 @tailwind utilities;
 
 input.ng-invalid.ng-touched {
-  @apply tce-border-red-600 dark:tce-border-red-700 tce-border-2 tce-m-[-1px]
+  @apply tce-border-red-600 tce-border-2 tce-m-[-1px]
 }
 
 .tce-error-box {
-  @apply tce-text-red-600 dark:tce-text-white dark:tce-bg-red-700 dark:tce-p-1 dark:tce-rounded dark:tce-border dark:tce-border-white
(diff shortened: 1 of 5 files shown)


src/styles.css
```diff
@@ -2,6 +2,14 @@
 @tailwind components;
 @tailwind utilities;
 
+input.ng-invalid.ng-touched {
+  @apply tce-border-red-600 dark:tce-border-red-700 tce-border-2 tce-m-[-1px]
```
jantoun-scottlogic, line 6: The darker border is a bit harder to see on dark mode but matches the background for the error message. I think that the border is a bit difficult to see either way on dark mode and the error message would be the main indication of an error so hopefully that's ok as long as the error message itself is clear.

mgriffin-scottlogic: My only thought is we might want to put the `dark:` parts into the `package-styles.css file` instead. Only because the SL site does not support dark mode, and you'll then get the red background if you have it set in your preferences:

image

jantoun-scottlogic: Done.

mgriffin-scottlogic (approved):

mgriffin-scottlogic (approved):

sdun-scottlogic:
Tested & Approved - SD.

PR #115, opened 2024-07-18
SFD 152 Add tabular data

jmain-scottlogic:
- Create a table to give an alternative display for the emissions data. Table follows treegrid pattern, https://www.w3.org/WAI/ARIA/apg/patterns/treegrid/
- Created a tab component
- Refactored carbon estimation to have 2 tabs, one for graph data and one for tabular data

TODO 
- [x] Confirm tab naming
    -  Updated to Diagram and Table
- [x] Confirm table styling
- [x] Add placeholder for tabular data


jmain-scottlogic:
Screen shot of the tab, with current names. Think the names could be better/clearer will make a post on teams to get clarity 
image

jmain-scottlogic:
Screen shot of the table. 
image

jmain-scottlogic:
Screen shot of table placeholder
image

src/app/carbon-estimation-table/carbon-estimation-table.component.html
```diff
@@ -0,0 +1,72 @@
...
+            tabindex="-1"
+            class="tce-flex tce-text-white tce-items-center"
+            (keydown.enter)="toggle(tableItem.category)">
+            <!-- eslint-disable-next-line @angular-eslint/template/click-events-have-key-events -->
+            <div
+              tabindex="-1"
+              (click)="toggle(tableItem.category)"
+              class="material-icons-outlined hover:tce-opacity-50 tce-m-1 tce-cursor-pointer">
+              {{ tableItem.expanded ? 'expand_less' : 'expand_more' }}
+            </div>
```
jantoun-scottlogic, line 39: There are quite a lot of accessibility pitfalls related to not using `<button>` elements to do this sort of thing. I think we should have a single `<button>` element that toggles the expansion.

jantoun-scottlogic:
> Screen shot of the table. image

Looks great overall but would it be possible to align the headings with the content a bit more? In particular the 'Emissions' looks a bit off relative to the data in the column.

src/app/carbon-estimation-table/carbon-estimation-table.component.html
```diff
@@ -0,0 +1,72 @@
...
+      <th class="tce-w-40">Emissions</th>
+    </tr>
+  </thead>
+  <tbody
+    (keydown.home)="homeEndKeyBoardEvent($event, true)"
+    (keydown.end)="homeEndKeyBoardEvent($event, false)"
+    (keydown.control.home)="homeEndKeyBoardEvent($event, true)"
+    (keydown.control.end)="homeEndKeyBoardEvent($event, false)"
+    (keydown.arrowup)="arrowKeyBoardEvent($event, 'up')"
+    (keydown.arrowdown)="arrowKeyBoardEvent($event, 'down')">
```
jantoun-scottlogic, line 15: Is it the fact that sections of the table collapse that necessitates custom handlers for keyboard interaction? Whilst I think the table looks great and works fine for mouse users this amount of programmatic handling of keystrokes and setting the tabindex looks like a red flag to me in terms of accessibility in the sense that I think we'll probably have fallen into a lot of traps on the way. There's also quite a lot of code to grok and work on if things do need to be changed. 

I'm wondering if the sections should really be collapsible. It's quite a nice feature for mouse users but there isn't that much information in the table when expanded so I don't think collapsing the table makes a huge difference in terms of parsing the data. Given that one of the main reasons for adding the table was to make the data more accessible I would be tempted to remove the collapsing section feature if that's the reason for the complexity.

src/app/carbon-estimation-table/carbon-estimation-table.component.html
jmain-scottlogic: I dont think the button is needed as it is following the [treegrid pattern](https://www.w3.org/WAI/ARIA/apg/patterns/treegrid/). So for keyboard users there are multiple ways to expand/collapse the row as part of the pattern which I think negates the potential pitfalls

jantoun-scottlogic: Ah ok, my bad. That's interesting. I still feel like it's strange to have one element for the mouse clicks and a separate one for the enter key and that the element that handles mouse clicks is not a button but I'll read the pattern more thoroughly.

jantoun-scottlogic: I think the div that handles the 'enter' key needs an aria label, at the moment the screen reader announces the 'expand_mode'/'expand_less' icon code which is ok but perhaps should be something like 'expand'/'collapse section'.

jantoun-scottlogic: Looking at the [example](https://www.w3.org/WAI/ARIA/apg/patterns/treegrid/examples/treegrid-1/) of the pattern, the arrow that handles the clicks isn't a button either and it doesn't seem you can navigate to it with the keyboard. The screen reader (NVDA) announces the cell entry (and the row and column numbers) when the cell that handles the 'enter' key is focussed rather than something like 'expand'/'collapse'. Their data is a little different to ours though.

jmain-scottlogic: I was thinking of going to the other way and making the div aria-hidden as it is just for visual users and keyboard users would use the row action or enter on cell.

jantoun-scottlogic: Yes, that might be a good idea for the arrow icon. It's also currently announced when focusing the row as a whole but that doesn't have the 'enter' key expansion behaviour. Given that it only interacts with the mouse perhaps it's supposed to be hidden from screen readers. The arrow icons in the [W3C example](https://www.w3.org/WAI/ARIA/apg/patterns/treegrid/examples/treegrid-1/) seems to be hidden from NVDA.

jantoun-scottlogic:
> Screen shot of table placeholder image

I'm wondering if we should remove the expand/collapse arrow icons or make it so you can't click them and add the disabled mouseover icon to the placeholder since they don't expand or collapse sections when clicked.

src/app/carbon-estimation-table/carbon-estimation-table.component.ts
```diff
@@ -0,0 +1,323 @@
...
+        emission.expanded = !emission.expanded;
+      }
+    });
+  }
+
+  public parentRowArrowKeyBoardEvent(
+    keyBoardEvent: Event,
+    direction: ArrowDirectionHorizontal,
+    parent: string,
+    expanded?: boolean
```
jantoun-scottlogic, line 65: What was the reason for making this argument optional?

src/app/carbon-estimation/carbon-estimation.component.ts
```diff
@@ -1,125 +1,79 @@
...
-  EmissionsLabels,
-  SVG,
-  getBaseChartOptions,
-  estimatorHeights,
-  tooltipFormatter,
-  placeholderData,
-  ApexChartSeriesItem,
-  ApexChartDataItem,
-} from './carbon-estimation.constants';
+import { ChangeDetectorRef, Component, computed, ElementRef, input, OnDestroy, OnInit, ViewChild } from '@angular/core';
```
jantoun-scottlogic, line 1:
```suggestion
import { ChangeDetectorRef, Component, ElementRef, input, OnDestroy, OnInit, ViewChild } from '@angular/core';
```
I think `computed` isn't being used now.

src/app/carbon-estimation/carbon-estimation.component.ts
```diff
@@ -1,125 +1,79 @@
...
 import { ExpansionPanelComponent } from '../expansion-panel/expansion-panel.component';
-import { Subscription, debounceTime, fromEvent } from 'rxjs';
+import { TabsComponent } from '../tab/tabs/tabs.component';
+import { TabItemComponent } from '../tab/tab-item/tab-item.component';
+import { CarbonEstimationTreemapComponent } from '../carbon-estimation-treemap/carbon-estimation-treemap.component';
+import { CarbonEstimation } from '../types/carbon-estimator';
+import { sumValues } from '../utils/number-object';
+import { estimatorHeights } from './carbon-estimation.constants';
+import { debounceTime, fromEvent, Observable, Subscription } from 'rxjs';
```
jantoun-scottlogic, line 9:
```suggestion
import { debounceTime, fromEvent, Subscription } from 'rxjs';
```
I don't think `Observable` is being used.

src/app/carbon-estimation/carbon-estimation.component.html
```diff
@@ -21,16 +21,15 @@ <h2 class="tce-text-2xl">Estimations</h2>
...
+    <tab-item [active]="true" [title]="'Diagram'" (tabSelected)="treemapSelected()">
+      <carbon-estimation-treemap
+        #treemap
+        [carbonEstimation]="carbonEstimation()"
+        [chartHeight]="chartHeight"></carbon-estimation-treemap>
+    </tab-item>
+    <tab-item [title]="'Table'">
+      <carbon-estimation-table [carbonEstimation]="carbonEstimation()"></carbon-estimation-table
+    ></tab-item>
+  </tabs>
```
jantoun-scottlogic, line 34: I think these should follow the [tabs pattern](https://www.w3.org/WAI/ARIA/apg/patterns/tabs/) which would require some changes in terms of keyboard interaction and roles. This could be looked into in a follow up ticket though.

src/app/tab/tabs/tabs.component.spec.ts
```diff
@@ -0,0 +1,41 @@
...
+import { TabItemComponent } from '../tab-item/tab-item.component';
+
+describe('TabsComponent', () => {
+  beforeEach(async () => {
+    await TestBed.configureTestingModule({
+      imports: [TabsComponent, TabItemComponent],
+    }).compileComponents();
+  });
+
+  it('should select 1st tab is no active tab', () => {
```
jantoun-scottlogic, line 13:
```suggestion
  it('should select 1st tab if no active tab', () => {
```

src/app/carbon-estimation-table/carbon-estimation-table.component.html
jantoun-scottlogic, line 1: In dark mode the text in the lighter coloured cells is white which is causing contrast issues.

image

src/app/carbon-estimation/carbon-estimation.constants.ts
```diff
@@ -1,10 +1,14 @@
...
 import { ChartOptions } from '../types/carbon-estimator';
 
 export enum EmissionsColours {
   Upstream = '#40798C',
+  UpstreamLight = '#bfdae2',
   Direct = '#CB3775',
   Indirect = '#91234C',
+  OperationLight = '#f2afd1',
   Downstream = '#4B7E56',
+  DownstreamLight = '#c1d9c3',
```
jantoun-scottlogic, line 11:
```suggestion
  Upstream = '#40798C',
  UpstreamLight = '#BFDAE2',
  Direct = '#CB3775',
  Indirect = '#91234C',
  OperationLight = '#F2AFD1',
  Downstream = '#4B7E56',
  DownstreamLight = '#C1D9C3',
```
Very minor nitpick but can we match the case in the new hex codes with the pre-existing ones?

src/app/carbon-estimation-table/carbon-estimation-table.component.ts
jmain-scottlogic: I think it was left over from a refactor. Will remove optionality as it is needed.

src/app/services/carbon-estimation-util.service.ts
```diff
@@ -0,0 +1,76 @@
...
+})
+export class CarbonEstimationUtilService {
+  constructor() {}
+
+  public getOverallPercentageLabel = (emissions: NumberObject): string => {
+    const percentage = sumValues(emissions);
+    return this.getPercentageLabel(percentage);
+  };
+
+  public getPercentageLabel = (percentage: number): string => (percentage < 1 ? '<1%' : Math.round(percentage) + '%');
```
mgriffin-scottlogic, line 17: Not introduced by this change but noticed this is a dupe of `tooltipFormatter` from `carbon-estimation.constants` (or vice versa). The treemap usage of `tooltipFormatter` could switch to using this from the service but I'm not quite sure what would be best to do with the custom tooltip and data label (unless this service provides them too?). It does seem good that these are tested in isolation.

jmain-scottlogic:
New empty table screenshot
image

src/app/services/carbon-estimation-util.service.ts
jmain-scottlogic: I can change this to be 
`public getPercentageLabel = tooltipFormatter;`
Then only one occurrence of the function and it will be covered by util service unit tests.

src/app/carbon-estimation/carbon-estimation.component.html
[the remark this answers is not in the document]
jmain-scottlogic: Yeah that sounds like a good plan. Created the following ticket, https://scottlogic.atlassian.net/browse/SFD-181

src/app/tab/tabs/tabs.component.ts
```diff
@@ -0,0 +1,28 @@
...
+      this.selectTab(this.tabs.first);
+    } else if (activeTabs.length > 1) {
+      this.selectTab(activeTabs[0]);
+    }
+  }
+
+  public selectTab(selectedTab: TabItemComponent) {
+    this.tabs.filter(tab => tab.active()).forEach(tab => tab.active.set(false));
+    selectedTab.active.set(true);
+    selectedTab.tabSelected.emit();
```
mgriffin-scottlogic, line 26: I thought this reaching into the tab item to emit an event was a little unexpected. I wonder if it might be a good use case for adding an `effect()` on the tab item that responds to the `active()` signal changes and emits when true?

src/app/carbon-estimation/carbon-estimation.component.ts
```diff
@@ -1,125 +1,79 @@
...
+  private hasResized = true;
+  private hasEstimationUpdated = false;
+  private carbonEstimationSubscription?: Subscription;
 
-  @ViewChild('chart') chart: ChartComponent | undefined;
-  @ViewChild('detailsPanel', { static: true, read: ElementRef }) detailsPanel!: ElementRef;
-
-  constructor(private changeDetectorRef: ChangeDetectorRef) {}
+  constructor(private changeDetectorRef: ChangeDetectorRef) {
+    this.carbonEstimationSubscription = toObservable(this.carbonEstimation).subscribe(() => {
```
mgriffin-scottlogic, line 42: I don't know if this would this be simpler with an `effect()` but it might feel a bit odd calling the signal just for the purposes of receiving updates.

```typescript
effect(() => {
  this.carbonEstimation();
  this.hasEstimationUpdated = true;
});
```

jmain-scottlogic: I do think it is cleaner than Observables, dont have to deal with the destroying the observable. Yeah slightly annoying you cant define the signal you want it to depend on.

src/app/carbon-estimation/carbon-estimation.component.ts
```diff
@@ -53,73 +55,23 @@ export class CarbonEstimationComponent implements OnInit, OnDestroy {
...
-  private getEmissionMadeUp(emission: ApexChartDataItem[]): string {
-    if (emission.length === 0) {
-      return '.';
+  public treemapSelected(): void {
+    if (this.hasResized || this.hasEstimationUpdated) {
+      this.hasResized = false;
+      this.hasEstimationUpdated = false;
+      this.changeDetectorRef.detectChanges();
+      this.treemap.chart?.updateOptions({});
     }
```
jantoun-scottlogic, line 74:
```suggestion
  public treemapSelected(): void {
    this.changeDetectorRef.detectChanges();
    this.treemap.chart?.updateOptions({});
```
Do we need the conditional logic here? Could this be simplified to the above?

jmain-scottlogic: I would be keen to keep the conditional logic, as it stops unnecessary re-rendering of the treemap

mgriffin-scottlogic: Out of interest I did a quick profile of switching between the tabs with and without the conditional:
image
vs
image
I didn't see much difference visibly but it's obviously doing more work than needed.

src/app/carbon-estimation-table/carbon-estimation-table.component.spec.ts
```diff
@@ -0,0 +1,357 @@
...
+  it('should set child emissions to display by default', () => {
+    component.tableData().forEach(emission => {
+      if (emission.level === 2 && emission.parent) {
+        expect(emission.display).toBeTrue();
+      }
+    });
+  });
+
+  it('should get emissions when getEmissions called', () => {
+    const emissions = component.getTableData(component.carbonEstimation());
```
mgriffin-scottlogic, line 70: Seems like `getTableData()` is only exposed for this test, maybe it should be private and test the `tableData` instead?

src/app/carbon-estimation-table/carbon-estimation-table.component.html
```diff
@@ -0,0 +1,81 @@
...
+<table role="treegrid" aria-label="Estimation of emissions" class="tce-table-fixed tce-w-full">
+  <thead>
+    <tr class="tce-border-b-2">
+      <th class="tce-text-left">Category</th>
+      <th class="tce-w-40 tce-text-left">Emissions</th>
+    </tr>
+  </thead>
+  @if (carbonEstimation()) {
+    <tbody
+      (keydown.home)="homeEndKeyBoardEvent($event, true)"
```
mgriffin-scottlogic, line 11: Could these functions pull the key information out of `KeyboardEvent.key`? These seem like they'd be easy to set up wrong, especially the booleans where you don't really have any context for what it means.

src/app/carbon-estimation-table/carbon-estimation-table.component.html
```diff
@@ -0,0 +1,81 @@
+<!-- eslint-disable @angular-eslint/template/interactive-supports-focus -->
+<table role="treegrid" aria-label="Estimation of emissions" class="tce-table-fixed tce-w-full">
+  <thead>
+    <tr class="tce-border-b-2">
+      <th class="tce-text-left">Category</th>
+      <th class="tce-w-40 tce-text-left">Emissions</th>
+    </tr>
+  </thead>
+  @if (carbonEstimation()) {
```
mgriffin-scottlogic, line 9: Worth checking `tableData().length > 0` instead?

src/app/carbon-estimation-table/carbon-estimation-table.component.ts
```diff
@@ -0,0 +1,310 @@
...
+  constructor(
+    private carbonEstimationUtilService: CarbonEstimationUtilService,
+    private changeDetector: ChangeDetectorRef
+  ) {}
+
+  public toggle(category: string): void {
+    this.tableData().forEach(emission => {
+      if (emission.level === 2 && emission.parent === category) {
+        emission.display = !emission.display;
+        this.expandedState[category] = emission.display;
```
mgriffin-scottlogic, line 53: Should the `expandedState` be set from the level 1 if statement? Here it's being updated up to 3 times.

mgriffin-scottlogic (dismissed):

src/app/carbon-estimation-table/carbon-estimation-table.component.html
jmain-scottlogic: Yeah they could, will update

src/app/carbon-estimation-table/carbon-estimation-table.component.html
jmain-scottlogic: Yeah it could. Will update as think it makes more sense as that is the table and tableData

src/app/carbon-estimation-table/carbon-estimation-table.component.ts
jmain-scottlogic: Will update

mgriffin-scottlogic (approved):

sdun-scottlogic:
Tested & Approved - SD.

PR #118, opened 2024-07-25
SFD-128: Add error summary for validation errors

jantoun-scottlogic:
[SFD-128 Jira ticket](https://scottlogic.atlassian.net/browse/SFD-128)

Adds an error summary to display form validation errors. Includes links to the invalid input fields.

Notes:
-   The 'calculate' button was previously disabled when there were validation errors. It has been enabled at all times but the error summary will be shown if there are validation errors. This is so that the error feedback is the same whether the form is submitted using the 'enter' key or the 'calculate' button. 

Error summary:
error-summary

Error summary dark mode:
error-summary-dark-mode

### DIFF
--- a/src/app/carbon-estimator-form/carbon-estimator-form.component.html
+++ b/src/app/carbon-estimator-form/carbon-estimator-form.component.html
@@ -1,7 +1,7 @@
 <form (ngSubmit)="handleSubmit()" [formGroup]="estimatorForm" class="tce-w-full tce-flex tce-flex-col tce-gap-6">
   <div formGroupName="upstream">
     @if (showErrorSummary) {
-      <error-summary [validationErrors]="getValidationErrors()"></error-summary>
+      <error-summary [validationErrors]="validationErrors"></error-summary>
     }
 
     <ng-container *ngTemplateOutlet="sectionHeader; context: formContext.upstream"></ng-container>
--- a/src/app/carbon-estimator-form/carbon-estimator-form.component.html
+++ b/src/app/carbon-estimator-form/carbon-estimator-form.component.html
@@ -85,7 +85,7 @@
           required
           [attr.aria-describedby]="(numberOfServers | invalidated) ? 'numberOfServersError' : null" />
         @if (numberOfServers | invalidated) {
-          <div class="tce-error-box tce-flex tce-gap-1" aria-live="polite" id="numberOfServersError">
+          <div class="tce-error-box tce-flex tce-items-center tce-gap-1" aria-live="polite" id="numberOfServersError">
             <span class="material-icons-outlined">error</span>
             <p>{{ errorConfig.numberOfServers.errorMessage }}</p>
           </div>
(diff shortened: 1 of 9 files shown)


src/app/carbon-estimator-form/carbon-estimator-form.component.html
```diff
@@ -14,9 +18,9 @@
           required
           [attr.aria-describedby]="(headCount | invalidated) ? 'headCountError' : null" />
         @if (headCount | invalidated) {
-          <div class="tce-error-box tce-flex tce-gap-1" aria-live="polite" id="headCountError">
+          <div class="tce-error-box tce-flex tce-items-center tce-gap-1" aria-live="polite" id="headCountError">
```
jmain-scottlogic, line 21: Should this class be added to the other errors?

src/app/error-summary/error-summary.component.html
```diff
@@ -0,0 +1,10 @@
+<div #errorSummary tabindex="-1" class="tce-error-summary tce-border-4 tce-rounded tce-p-3 tce-mt-2 tce-mb-5">
+  <h2 class="tce-text-2xl"><strong>There is a problem</strong></h2>
```
jmain-scottlogic, line 2: Think adding a ! might look good

src/app/carbon-estimator-form/carbon-estimator-form.component.html
jantoun-scottlogic: Yes that seems like a good idea. I didn't add it initially because it was causing issues when some of the flex boxes contained anchor tags but they've been removed now so I've added that class to all the error boxes.

src/app/error-summary/error-summary.component.html
jantoun-scottlogic: Yes that might be good. I'll put some screenshots with and without the icon in Teams and ask for some feedback.

jantoun-scottlogic: It looks like on balance going without the '!' was preferred.

src/app/carbon-estimator-form/carbon-estimator-form.component.ts
```diff
@@ -20,6 +21,26 @@ const locationDescriptions: Record<WorldLocation, string> = {
   'LATIN AMERICA AND CARIBBEAN': 'in Latin America or the Caribbean',
 };
 
+export type ValidationError = {
+  inputId: string;
+  errorMessage: string;
+};
+
+const errorConfig = {
```
mgriffin-scottlogic, line 29: Maybe put this in the `.constants.ts` file?

Would also feel better to have the ValidationError in a separate file than having a circular reference.

jmain-scottlogic (approved):

mgriffin-scottlogic (approved):

src/app/carbon-estimator-form/carbon-estimator-form.component.ts
jantoun-scottlogic: Sounds good. Should I move the `locationDescriptions` to the `.constants.ts` file as well?

mgriffin-scottlogic: Yeah, that seems like a good idea. I'd forgotten there were two things with the same name in the same file.

mgriffin-scottlogic (approved):

sdun-scottlogic:
Tested & Approved - SD.

PR #121, opened 2024-08-02
SFD-171: Persist form state

jantoun-scottlogic:
[SFD-171 Jira ticket](https://scottlogic.atlassian.net/browse/SFD-171)

Persists the form state in session storage when swapping away from the form view or navigating away from the page.

### DIFF
--- a/src/app/carbon-estimator-form/carbon-estimator-form.component.spec.ts
+++ b/src/app/carbon-estimator-form/carbon-estimator-form.component.spec.ts
@@ -7,7 +7,7 @@ class MockStorageService {
   storage = new Map<string, string>();
 
   get(key: string): string | null {
-    return this.storage.get(key) || null;
+    return this.storage.get(key) ?? null;
   }
 
   set(key: string, value: string): void {
--- a/src/app/carbon-estimator-form/carbon-estimator-form.component.spec.ts
+++ b/src/app/carbon-estimator-form/carbon-estimator-form.component.spec.ts
@@ -13,6 +13,10 @@ class MockStorageService {
   set(key: string, value: string): void {
     this.storage[key] = value;
   }
+
+  removeItem(key: string): void {
+    delete this.storage[key];
+  }
 }
 
 describe('CarbonEstimatorFormComponent', () => {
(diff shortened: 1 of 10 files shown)


src/app/carbon-estimator-form/carbon-estimator-form.component.spec.ts
```diff
@@ -18,7 +32,6 @@ describe('CarbonEstimatorFormComponent', () => {
 
   it('should create component form in a valid state', () => {
     expect(component).toBeTruthy();
-    component.ngOnInit();
```
jantoun-scottlogic, line 21: I don't think we need these explicit calls because `ngOnInit` is being called in the `beforeEach` block.

src/app/carbon-estimator-form/carbon-estimator-form.component.ts
```diff
@@ -187,6 +227,7 @@ export class CarbonEstimatorFormComponent implements OnInit {
 
   public resetForm() {
     this.estimatorForm.reset();
+    this.storeFormState();
```
jantoun-scottlogic, line 235: We could add some extra functionality to the storage service and clear the stored state here but I went with just storing it for now since it's slightly simpler.

jantoun-scottlogic: I decided to update this to clear the stored state.

src/app/services/storage.service.ts
```diff
@@ -0,0 +1,25 @@
+import { Inject, Injectable, InjectionToken } from '@angular/core';
+
+export const BROWSER_STORAGE = new InjectionToken<Storage>('Browser Storage', {
```
jantoun-scottlogic, line 3: I followed the [Angular docs example](https://angular.dev/guide/di/di-in-action) here but we don't actually need this injection token at the moment. It might be useful for injecting mocks for testing in the future though so I left it in for now. (The tests for the estimator form mock the `StorageService` class.)

jmain-scottlogic (approved):

sdun-scottlogic:
Tested & Approved - SD.

PR #127, opened 2024-08-30
SFD-200: Persist diagram display

mgriffin-scottlogic:
[SFD-200 Jira ticket](https://scottlogic.atlassian.net/browse/SFD-200)

## Description of Change

Restore the diagram state when navigating away from the site and returning.

Possibly one for @jantoun-scottlogic to review the approach, as it seemed simplest to just call `handleSubmit()` if it was done previously, instead of having the diagram store state as well. There are some edge cases, where the form has been changed but not resubmitted before navigating away and it also does not restore which display tab was active when looking at the table.

I also didn't need to update any tests to make this change but I wasn't sure whether to get into checking what the form storage service was called with and how the init process changes when existing state is found.

## Checklist

### DIFF
--- a/src/app/carbon-estimator-form/carbon-estimator-form.component.ts
+++ b/src/app/carbon-estimator-form/carbon-estimator-form.component.ts
@@ -98,6 +98,7 @@ export class CarbonEstimatorFormComponent implements OnInit, OnDestroy {
     showErrorSummary: false,
     validationErrors: [],
   };
+  public submitted = false;
 
   public compareCostRanges = compareCostRanges;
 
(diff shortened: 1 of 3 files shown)


jantoun-scottlogic:
I think this is fine. It would be nice to resolve the edge cases but I think that would take a bit of work. I've created [SFD-208](https://scottlogic.atlassian.net/browse/SFD-208) to look into it.

jantoun-scottlogic (approved):

sdun-scottlogic:
Tested & Approved - SD.

PR #129, opened 2024-09-05
SFD-194: Fix error on app startup

mgriffin-scottlogic:
[SFD-194 Jira ticket](https://scottlogic.atlassian.net/browse/SFD-194)

## Description of Change

Ensure internal apex chart is ready before attempting to update its options.

There is a short period of time where the ng-apexchart `ChartComponent` has been created but it hasn't finished setting up the ApexChart element it's wrapping. Potentially a bug in their code but the `chartReady` event suggests it's something to be expected.

## Checklist


jcamilleri-scottlogic (approved):

sdun-scottlogic:
Tested & Approved - SD.

PR #130, opened 2024-09-11
SFD-193: Avoid duplication of build and test steps in publish pipeline

mgriffin-scottlogic:
[SFD-193 Jira ticket](https://scottlogic.atlassian.net/browse/SFD-193)

## Description of Change

Add a shared workflow to publish pages, which can only be called from other workflows. There are now 3 workflows that can be triggered manually:

- **Build and Test** (`buildAndTest.yml`)
- **Deploy estimator pages** (`pages.yml`)
- **Publish new package** (`publish-package.yml`)

And one that can only be called from others:

- **Shared Pages publish workflow** (`shared-pages.yml`)

**Deploy estimator pages** is now a simple workflow that only calls other workflows, `buildAndTest` followed by `shared-pages`.  In theory we should not need to use this in future but we will keep it for now. **Publish new package** has only changed to call `shared-pages`.

### DIFF
--- a/.github/workflows/pages.yml
+++ b/.github/workflows/pages.yml
@@ -3,47 +3,16 @@ name: Deploy estimator Pages
 on:
   # Allows you to run this workflow manually from the Actions tab
   workflow_dispatch:
-  workflow_call:
-
-# Sets permissions of the GITHUB_TOKEN to allow deployment to GitHub Pages
-permissions:
-  contents: read
-  pages: write
-  id-token: write
-
-# Allow only one concurrent deployment, skipping runs queued between the run in-progress and latest queued.
-# However, do NOT cancel in-progress runs as we want to allow these production deployments to complete.
-concurrency:
-  group: 'pages'
-  cancel-in-progress: false
 
 jobs:
-  # Build job
-  build_page:
-    runs-on: ubuntu-latest
-    steps:
-      - name: Check out code
-        uses: actions/checkout@v4
-      - name: Set up environment
-        uses: ./.github/actions/set-up-environment
-      - name: Build app
-        run: npm run build
-      - name: Upload artifact
-        uses: actions/upload-pages-artifact@v3
-        with:
-          path: ./dist/tech-carbon-estimator
-
   build_and_test:
     uses: ./.github/workflows/buildAndTest.yml
-    needs: build_page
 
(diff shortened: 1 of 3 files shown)


.github/workflows/pages.yml
cfisher-scottlogic, line 1: Do we still need this workflow? My understanding is that this can only be triggered manually, and it's basically a call to `shared pages`. Could we just make `shared-pages` manually-triggerable and remove this one?

.github/workflows/shared-pages.yml
cfisher-scottlogic, line 1: I'm a bit far removed from this one, but I suppose it's worth me asking anyway. What makes the pages `shared`, compared opposed to the `pages` we had before? Perhaps the real question I'm asking here is what are GitHub pages, and why are we deploying to them? 😅

.github/workflows/publish-package.yml
cfisher-scottlogic, line 1: How is this one triggered?

.github/workflows/pages.yml
mgriffin-scottlogic: I think in the long run we shouldn't need this workflow, I'm leaving it in case there are any issues with the `publish-package` workflow and we still need to publish the pages separately. The `shared-pages` workflow I want to think of as more like a composite action (but I don't think you can have multiple jobs in those), keeping the build and test part out of it (which caused the duplicate artifact problem). This one then means you can never publish the pages manually without the tests passing.

.github/workflows/shared-pages.yml
mgriffin-scottlogic: So GitHub pages is some web hosting that you can use from any repo, which we're using for the estimator that we include on the [Scott Logic site](https://www.scottlogic.com/technology-carbon-estimator). Because it was easier to include the estimator via an iframe on that site, we have it point to the pages url for this repo, which is https://scottlogic.github.io/sl-tech-carbon-estimator/. That way whenever we publish a new npm package, which the [Technology-Carbon-Standard](https://github.com/ScottLogic/Technology-Carbon-Standard) repo can use, we're also updating the Scott Logic site to use the latest version.

The `pages` workflow existed before any of this and we initially thought it was going to be more for testing but has morphed into being how we host it partially.

cfisher-scottlogic (approved):

PR #133, opened 2024-10-03
SFD-181: Implement aria tabs patterns

jantoun-scottlogic:
Resolves #&lt;Issue number&gt;

[&lt;Issue Code&gt; Jira ticket](https://scottlogic.atlassian.net/browse/Issue-Code)

## Description of Change

Write a brief but concise description of the changes you have made.

Include any relevant screenshots of visual changes.

## Checklist

### DIFF
--- a/src/app/tech-carbon-estimator/tech-carbon-estimator.component.html
+++ b/src/app/tech-carbon-estimator/tech-carbon-estimator.component.html
@@ -11,7 +11,7 @@
             (formReset)="handleFormReset()"></carbon-estimator-form>
         </tab-item>
         <tab-item [title]="'Assumptions and Limitations'">
-          <assumptions-and-limitation aria-live="polite"></assumptions-and-limitation>
+          <assumptions-and-limitation></assumptions-and-limitation>
         </tab-item>
       </tabs>
     </div>
--- a/src/app/tab/tab-item/tab-item.component.html
+++ b/src/app/tab/tab-item/tab-item.component.html
@@ -1,3 +1,8 @@
-@if (active()) {
+<div
+  role="tabpanel"
+  aria-live="off"
+  [hidden]="!active()"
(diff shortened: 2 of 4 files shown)


playwright/tests/test_17_EstTool_UI_AssumpLimitations.py
```diff
@@ -92,7 +92,7 @@ def test_example(page: Page) -> None:
     expect(page.get_by_text("Clearly there is a large")).to_be_visible()
     expect(page.get_by_text("The estimated kWh of cloud")).to_be_visible()
     expect(page.get_by_role("heading", name="Downstream Emissions")).to_be_visible()
-    expect(page.get_by_text("At present we focus on the")).to_be_visible()
+    expect(page.get_by_text("To do this we have collated")).to_be_visible()
```
jantoun-scottlogic, line 95: Now that the panels are hidden when not selected there are 2 elements in the DOM with the old text so I've used the start of the second sentence in the section instead.

src/app/tech-carbon-estimator/tech-carbon-estimator.component.html
```diff
@@ -11,7 +11,7 @@ <h1 class="tce-text-3xl tce-mb-6">Technology Carbon Estimator</h1>
             (formReset)="handleFormReset()"></carbon-estimator-form>
         </tab-item>
         <tab-item [title]="'Assumptions and Limitations'">
-          <assumptions-and-limitation aria-live="polite"></assumptions-and-limitation>
```
jantoun-scottlogic, line 14: I originally removed this when changing some other aria-live values but having reverted those changes and rechecked, I don't think this is changing the behaviour although that could just be my screen reader settings. I've left it off for now since other elements placed directly inside the tab-item components do not have this set.

PR #164, opened 2025-07-29
#141 Review Dependencies and Vulnerabilities

### LINKED ISSUE #141: [TECH DEBT] Review dependencies, vulnerabilities and additional security
Review and fix vulnerabilities currently on main/develop branches.
- https://github.com/ScottLogic/sl-tech-carbon-estimator/security/dependabot
- NPM Audit:
  -27 vulnerabilities (6 low, 9 moderate, 10 high, 2 critical)

Review Guthub security settings and enable additional features if we feel like they may be beneficial:
- Code Security

dmoorhouse-scottlogic:
Should resolve #141 

## Description of Change

Updates npm packages and resolves breaking changes

## Checklist


src/app/carbon-estimation/carbon-estimation.constants.ts
```diff
@@ -82,9 +82,7 @@ export const getBaseChartOptions = (isPlaceholder: boolean) => {
       fontFamily: 'ui-sans-serif, system-ui, sans-serif',
       fontWeight: '400',
       markers: {
-        height: 20,
-        width: 10,
-        radius: 2,
+        strokeWidth: 2,
```
mgriffin-scottlogic, line 85: Were these just changes in available options?

mgriffin-scottlogic (commented):
Only other question looking at the commits, did you manually update the package-lock.json to fix the koa vulnerability? Generally preferred that we don't touch that and use the overrides section of package.json to ensure that problematic versions are ignored if the lock has to be re-generated for any reason.

dmoorhouse-scottlogic:
> Only other question looking at the commits, did you manually update the package-lock.json to fix the koa vulnerability? Generally preferred that we don't touch that and use the overrides section of package.json to ensure that problematic versions are ignored if the lock has to be re-generated for any reason.

@mgriffin-scottlogic No this was an audit change that updated the package-lock

mgriffin-scottlogic:
> If we want to merge this in to develop first and then resolve the tailwind dependency, we will need to override those checks. Apparently this is an option in the branch protection rules, not sure if you have visibility on this @mgriffin-scottlogic but I don't seem to.

I'm not sure I can see anything different, I think the only option is the bypass at the top, which applies to everything in the ruleset.

image

We could temporarily suspend the playwright checks (it will still run them), if they're likely to change a few times in quick succession.

PR #188, opened 2025-08-08
175 testing optimise tests by moving repeated assertions into helper methodsfiles

### LINKED ISSUE #175: [TESTING] Optimise tests by moving repeated assertions into helper methods/files
Each of our e2e tests has many repeated lines.
These are mostly visibility checks i.e "expect this button to be visible before continuation of test"
We can extract these into a separate file to reduce the overall size of the test.

mnyamunda-scottlogic:
Resolves #175 

## Description of Change

Optimising e2e tests by extracting repeated steps into helper files 

## Checklist

### DIFF
--- a/e2e-tests/test-helpers.ts
+++ b/e2e-tests/test-helpers.ts
@@ -12,18 +12,17 @@ export async function gotoHome(page: Page) {
 }
 
 export async function expansionPanelClick(page: Page, panelName: string, action: 'Show details' | 'Hide details') {
-  await expect(page.locator('expansion-panel').filter({ hasText: panelName }).getByLabel(action)).toBeVisible();
   await page.locator('expansion-panel').filter({ hasText: panelName }).getByLabel(action).click();
 }
 
-export async function defaultPageElementVisibility(page: Page) {
-  await organisationVisibility(page);
-  await onPremiseVisibility(page);
-  await cloudVisibility(page);
-  await endUserVisibility(page);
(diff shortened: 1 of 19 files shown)


e2e-tests/test-helpers.ts
kellis-sl, line 1: As we're abstracting away the assertions here (the 'expect' statements), I'd suggest that the helper methods are named so that it's still clear what they are doing to someone who reads just the spec file and doesn't want to go look at the helper method

i.e.`assertOrganisationSectionVisibility`

e2e-tests/test-helpers.ts
```diff
@@ -0,0 +1,80 @@
...
+  expect(results.violations).toEqual([]);
+};
+
+export async function gotoHome(page: Page) {
+  await page.goto('/');
+  await expect(page.getByRole('heading', { name: 'Carbon Estimator' })).toBeVisible();
+}
+
+export async function expansionPanelClick(page: Page, panelName: string, action: 'Show details' | 'Hide details') {
+  await expect(page.locator('expansion-panel').filter({ hasText: panelName }).getByLabel(action)).toBeVisible();
```
kellis-sl, line 15: I would say do don't specifically need this, if the following step can't find the locator to click on, it will throw an error anyway

kellis-sl (changes requested):
Couple of suggested changes

kellis-sl (approved):

PR #189, opened 2025-08-08
139 feature implement ability to display the co2 values as well as percentages

bstinchcombe-scottlogic:
Resolves #&lt;Issue number&gt;

[&lt;Issue Code&gt; Jira ticket](https://scottlogic.atlassian.net/browse/Issue-Code)

## Description of Change

Write a brief but concise description of the changes you have made.

Include any relevant screenshots of visual changes.

## Checklist


src/styles.css
```diff
@@ -36,3 +36,64 @@ input.ng-invalid.ng-touched {
 .tce-active-tab, .tce-active-tab:hover {
   @apply tce-cursor-default tce-border-sky-800
 }
+
+.tce-totalchart-header {
```
mgriffin-scottlogic, line 40: On some of these, I don't know if we want to go down the route of adding more classes with regular CSS over the tailwind ones? Makes it a bit inconsistent across the whole project. Generally we've only added classes of our own when we anticipate they will be used in multiple places or will need to be modified in the package vs Scott Logic branded version.

bstinchcombe-scottlogic: Have updated these to use Tailwind classes

src/styles.css
```diff
@@ -87,6 +88,11 @@ input.ng-invalid.ng-touched {
   display: none;
 }
 
+.tce-chart-figure {
+  margin-left: 0px;
```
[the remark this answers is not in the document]
bstinchcombe-scottlogic: Looks like this is still needed when checking locally

src/app/services/carbon-estimation.service.spec.ts
```diff
@@ -57,21 +57,46 @@ function expectPartialNumberCloseTo(actual: NumberObject, expected: NumberObject
   }
 }
 
+// function expectPartialEstimationCloseTo(
```
mgriffin-scottlogic, line 60: Some temp stuff to be cleaned up here?

I'd forgotten how complicated this method was, and it kind of seems like it's getting more so now that it can deal with two types of object. Is it worth leaving this as specifically for percentages since we haven't added new tests on the values?

Also feels like it could now silently avoid errors as it will just skip over keys you expected but are not part of the actual, which doesn't seem right. Maybe assert that it is present before trying to use it?

src/app/carbon-estimation-table/carbon-estimation-table.component.html
```diff
@@ -1,11 +1,10 @@
 <!-- eslint-disable @angular-eslint/template/interactive-supports-focus -->
 <table role="treegrid" aria-label="Estimation of emissions" class="tce:table-fixed tce:w-full">
   <thead>
-    <tr class="tce-border-b-2">
-      <th class="tce-text-left">Emissions Category</th>
-      <th class="tce-w-40 tce-text-left">Estimated CO<sub>2</sub> E</th>
-      <th class="tce-w-40 tce-text-left">% of Total Estimate</th>
-    </tr>
+    <tr class="tce:border-b-2">
```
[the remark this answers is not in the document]
bstinchcombe-scottlogic: replaced

dmoorhouse-scottlogic (approved):

Issue #182, opened 2025-08-07
[FEATURE] CO2 Library Abstraction Layer


dmoorhouse-scottlogic:
**Description**
Recent testing highlighted the need for a wrapper for a CO2 library. 
This would:

- Allow swap out of CO2 library implementation
- Mock library for unit testing

**Acceptance Criteria**

- Given a test environment during unit testing, a mock library is returned from the abstraction
- Given a real environment, the CO2.js library is return from the abstraction

PR #207, opened 2025-08-15
#182 Add CO2 library abstraction

dmoorhouse-scottlogic:
## Description of Change

feat: Add wrapper around CO2 library to isolate testing issues during node package updates

- Added ICO2Calculator wrapper
- Added Factory to return appropriate implementation
- Added environment variables 

## Checklist

### DIFF
--- a/angular.json
+++ b/angular.json
@@ -38,12 +38,32 @@
                   "maximumError": "4kb"
                 }
               ],
-              "baseHref": "/sl-tech-carbon-estimator/"
+              "baseHref": "/sl-tech-carbon-estimator/",
+              "fileReplacements": [
+                {
+                  "replace": "src/environments/environment.ts",
+                  "with": "src/environments/environment.prod.ts"
+                }
+              ]
             },
             "npm-package": {
               "index": {"input": "src/package-index.html", "output": "index.html"},
               "assets": [],
-              "styles": ["src/styles.css", "src/package-styles.css"]
+              "styles": ["src/styles.css", "src/package-styles.css"],
+              "fileReplacements": [
+                {
+                  "replace": "src/environments/environment.ts",
+                  "with": "src/environments/environment.prod.ts"
+                }
+              ]
+            },
+            "fake-object": {
+              "fileReplacements": [
+                {
+                  "replace": "src/environments/environment.ts",
+                  "with": "src/environments/environment.fake-object.ts"
+                }
+              ]
             },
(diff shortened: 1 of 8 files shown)


kellis-sl (commented):
Overall this is looking good.

Have discussed with @dmoorhouse-scottlogic about the possibility of moving hardcode test values to constants or a helper file of some sort, which he agrees makes sense. Once done, we then just need to decide on the test values.

src/app/estimation/estimate-downstream-emissions.spec.ts
```diff
@@ -1,9 +1,22 @@
...
+    TestBed.configureTestingModule({
+      declarations: [estimateDownstreamEmissions],
+      providers: [
+        { provide: CO2_CALCULATOR, useValue: new FakeCO2Calculator('object')}
+      ]
+    })
+  })
+
   const carbonIntensity = 500;
+  const co2Calc = TestBed.inject(CO2_CALCULATOR);
```
kellis-sl, line 19: This is declared both outside and inside a test, is that right,hould it just be inside the test?

src/app/estimation/estimate-downstream-emissions.spec.ts
```diff
@@ -1,9 +1,22 @@
...
 import { Downstream, DownstreamEstimation, PurposeOfSite, basePurposeArray } from '../types/carbon-estimator';
 import { sumValues } from '../utils/number-object';
 import { estimateDownstreamEmissions } from './estimate-downstream-emissions';
+import { CO2_CALCULATOR } from '../facades/CO2InjectionToken';
+import { FakeCO2Calculator } from '../facades/FakeCO2Calculator';
 
 describe('estimateDownstreamEmissions()', () => {
+  beforeEach(() => {
+    TestBed.configureTestingModule({
+      declarations: [estimateDownstreamEmissions],
```
kellis-sl, line 11: I don't think functions should/can be in the declarations array. I think this may be why the dependency injection isn't working correctly and hence tests are passing? I think this needs to go in a `providers` array based on some testBed examples I just looked at.

kellis-sl: On second look at this, I don't think this is needed in declartions at all. You import the function and then call it inside the tests when needed so doesn't need to be part of the testBed config

kellis-sl (approved):

PR #214, opened 2025-08-19
Audit npm packages

kellis-sl:
Fix npm vulnerabilities


kellis-sl:
These were fixed with `npm audit` so I think we're good?

PR #226, opened 2025-08-29
Test/implement pom

### LINKED ISSUE #215: [TESTING] Test suite POM conversion
Convert current test suite to use Page Object Model

mnyamunda-scottlogic:
Resolves #215 

## Description of Change

- Separate tests out into different page object 
- Refactor e2e tests with new pom properties/methods
- Remove redundant comments

## Checklist

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3332,14 +3332,15 @@
       }
     },
     "node_modules/@inquirer/editor": {
-      "version": "4.2.15",
-      "resolved": "https://registry.npmjs.org/@inquirer/editor/-/editor-4.2.15.tgz",
-      "integrity": "sha512-wst31XT8DnGOSS4nNJDIklGKnf+8shuauVrWzgKegWUe28zfCftcWZ2vktGdzJgcylWSS2SrDnYUb6alZcwnCQ==",
+      "version": "4.2.17",
+      "resolved": "https://registry.npmjs.org/@inquirer/editor/-/editor-4.2.17.tgz",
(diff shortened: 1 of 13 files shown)


e2e-tests/page-objects/table-section.ts
```diff
@@ -0,0 +1,84 @@
...
+    this.indirectEmissionsEstimate = page.getByRole('gridcell', { name: 'Indirect Emissions Estimate', exact: true });
+    this.cloudServices = page.getByRole('gridcell', { name: 'Cloud Services', exact: true });
+    this.downstreamEmissionsEstimate = page.getByRole('gridcell', {
+      name: 'Downstream Emissions Estimate',
+      exact: true,
+    });
+    this.endUserDevices = page.getByRole('gridcell', { name: 'End-User Devices', exact: true });
+    this.networkDataTransfer = page.getByRole('gridcell', { name: 'Network Data Transfer', exact: true });
+    this.totalEmissionsEstimate = page.getByRole('gridcell', { name: 'Total Emissions Estimate', exact: true });
+    this.kilogramsColumn = page.locator('td:nth-child(2)');
```
mgriffin-scottlogic, line 50: Given there are tables in the assumptions section, I think these should be more specific. Either check they are visible or that they are children of the carbon-estimation-table element (Possibly not ideal but I'd hope the element name should be pretty stable, there's not really anything else I can think of to identify a unique parent).

mgriffin-scottlogic: Reading a little on Playwright's recommendations, the best way to identify it might be:

```suggestion
    this.kilogramsColumn = page.getByRole('treegrid').locator('td:nth-child(2)');
```

Which is something we added along with the keyboard support to open and close the sections.

kellis-sl (approved):

PR #229, opened 2025-09-03
140 feature implement ability to export data in some format

bstinchcombe-scottlogic:
feature 140 export options for JSON and PDF
This adds a drop down above the chart/table which gives three options
- export JSON, just the results, kg and %
- export JSON with inputs, same again but includes the values the user put into the estimator
- export PDF 2 opens a preview modal showing a 2 page document with customisable name  which displays the chart, table, the users inputs and the disclaimer provided in the estimator.

### DIFF
--- a/src/app/carbon-estimation/carbon-estimation.component.ts
+++ b/src/app/carbon-estimation/carbon-estimation.component.ts
@@ -117,6 +117,7 @@ export class CarbonEstimationComponent implements OnInit, OnDestroy {
     if (typeof exportObject === 'undefined') {
       return '';
     }
+
     const estimateJson = JSON.stringify(exportObject, null, 2);
     const blob = new Blob([estimateJson], { type: 'application/json' });
     const carbonEstimationJSONUrl = URL.createObjectURL(blob);


src/app/carbon-estimation/carbon-estimation.component.ts
```diff
@@ -113,7 +113,7 @@ export class CarbonEstimationComponent implements OnInit, OnDestroy {
     return this.getJSONExportUrl(exportObject);
   }
 
-  private getJSONExportUrl(exportObject: unknown): string {
+  private getJSONExportUrl(exportObject: CarbonEstimation|jsonExport|undefined): string {
     const estimateJson = JSON.stringify(exportObject, null, 2);
```
dmoorhouse-scottlogic, line 117: With the exportObject possibly being undefined is it worth adding checks before trying to stringify the value?

bstinchcombe-scottlogic: I think by disabling the export menu when no estimation data was present it wasn't actually possible at runtime for exportObject to be undefined but I've added the type check as that was probably a bit fragile

dmoorhouse-scottlogic (approved):

dmoorhouse-scottlogic (approved):

PR #235, opened 2025-09-16
Add automation to cover export scenarios

mnyamunda-scottlogic:
## Description of Change
 
- Export is disabled if no calculation is performed
- Export json download and assert values for both schema types, with and without inputs
- 
## Checklist


dmoorhouse-scottlogic (approved):
Looks good to me

PR #253, opened 2025-10-24
#219 Isolate styling with shadow dom

### LINKED ISSUE #219: [BUG] Latest version does not isolate Estimator styling
**Describe the bug**
When the Tech Carbon Standard is updated to use version `0.4.0`, the styling of the estimator is different and appears to be overridden by the TCS site styling.

**To Reproduce**
1. Clone the https://github.com/ScottLogic/Technology-Carbon-Standard repo
2. Update the `devDependencies` in its `package.json` to use `"@scottlogic/tech-carbon-estimator": "^0.4.0"`
3. Run `npm install` and `bundle install`
4. Run the site locally with `bundle exec jekyll serve`
5. Navigate to http://localhost:4000/estimator and view the estimator within the TCS site

**Expected behavior**
Estimator portion of the site appears as when testing the estimator locally. The styling of the rest of the TCS site should not be affected by including the estimator component.

**Screenshots**

Live view of TCS with sized headers and active tab bottom border

Image

TCS using updated package with header sizing and active tab affected

Image

Inspection showing bottom border being overridden

Image

Inspection showing main.css taking precedence

Image

**Additional context**
Some thoughts on possible issues/solutions here:

- TCS site is using older version of Tailwind, and therefore does not put everything into layers by default
- CSS rules without a layer take priority over those with, no matter the ordering of their inclusion
- Use of Shadow DOM for the estimator component seems like the proper solution here but there may be issues with getting this working with Tailwind via Angular.

robatwilliams:
After #230 and [attempt ](https://github.com/ScottLogic/sl-tech-carbon-estimator/tree/219-isolate-estimator-styling-flattened) to get it all working, we decided to try an alternative approach which keeps things simpler, and more "on-piste" in terms of build tooling.

Fixes #219

### DIFF
--- a/src/app/tech-carbon-estimator/tech-carbon-estimator.component.ts
+++ b/src/app/tech-carbon-estimator/tech-carbon-estimator.component.ts
@@ -9,7 +9,6 @@ import { AssumptionsAndLimitationComponent } from '../assumptions-and-limitation
 import { DisclaimerComponent } from '../disclaimer/disclaimer.component';
 import { TabsComponent } from '../tab/tabs/tabs.component';
 import { TabItemComponent } from '../tab/tab-item/tab-item.component';
-import { style } from '@angular/animations';
 
 @Component({
   selector: 'tech-carbon-estimator',


README.md
```diff
@@ -22,7 +22,6 @@ You will need to import the following files to use the tech-carbon-estimator:
 
 - main.js
 - polyfill.js
-- styles.css
```
robatwilliams: Seemed likely this needed to be removed, but I haven't seen it used - Dan, I guess you will when you try embedding it in the website.

With the style isolation, we are changing our ask of consumers from "include this in your global styles somewhere" to "put this file at `/styles.css` for us". Which is very rigid.

Perhaps we should allow the styles url to be given as a component attribute - including any path and cache-busting suffixes.

Having the whole bundled styles within the js bundle (so we can import it as a string rather than from a URL path) would eliminate the issue altogether, but then we're back into custom build system/tooling territory.

Why is it necessary to have separate sl-style and package-style? If we could have package-style be always used, we could specify it in `styleUrl` and simplify things a lot, and then do something else (e.g. style tag) for adaptations the demo page requires.

src/app/tech-carbon-estimator/tech-carbon-estimator.component.ts
```diff
@@ -35,8 +38,34 @@ export class TechCarbonEstimatorComponent {
...
+    tailwindShadowDomStylesLink.rel = 'stylesheet';
+    tailwindShadowDomStylesLink.type = 'text/css';
+    tailwindShadowDomStylesLink.href = 'tailwind-shadowdom-styles.css'; // This is the `bundleName` of the concatenated styles file
+
+    const stylesLink = document.createElement('link');
+    stylesLink.rel = 'stylesheet';
+    stylesLink.type = 'text/css';
+    stylesLink.href = 'styles.css'; // This is the `bundleName` of the concatenated styles file
+
+    const googleFontsLink = document.createElement('link');
```
robatwilliams: Now that we have 3 links to insert, a utility method would reduce repetition

src/app/tech-carbon-estimator/tech-carbon-estimator.component.ts
```diff
@@ -35,8 +38,34 @@ export class TechCarbonEstimatorComponent {
...
+  ) {
+    this.insertShadowStylesLink()
+  }
+
+  private insertShadowStylesLink() {
+    // Reasons for this approach:
+    // 1. Angular global injection would insert the tag in the page root, so we disabled it.
+    // 2. Component `styleUrl` wouldn't allow us to vary stylesheets based on build configurations.
+
+    const tailwindShadowDomStylesLink = document.createElement('link');
```
robatwilliams: This isn't being added to the DOM

src/app/tech-carbon-estimator/tech-carbon-estimator.component.ts
```diff
@@ -35,9 +39,46 @@ export class TechCarbonEstimatorComponent {
...
+      }
+    }
+
+    let stylesPath = assetPath + styleHref;
+
+    let stylesLink = document.createElement('link');
+    stylesLink.rel = 'stylesheet';
+    stylesLink.type = 'text/css';
+    stylesLink.href = stylesPath;
+    return stylesLink;
```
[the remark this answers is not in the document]
dmoorhouse-scottlogic: We return the link to insertShadowStyleLink which adds it to the shadow dom

src/app/tech-carbon-estimator/tech-carbon-estimator.component.ts
```diff
@@ -9,6 +9,7 @@ import { AssumptionsAndLimitationComponent } from '../assumptions-and-limitation
 import { DisclaimerComponent } from '../disclaimer/disclaimer.component';
 import { TabsComponent } from '../tab/tabs/tabs.component';
 import { TabItemComponent } from '../tab/tab-item/tab-item.component';
+import { style } from '@angular/animations';
```
kellis-sl: This isn't being used

src/app/carbon-estimation-treemap/carbon-estimation-treemap.component.html
```diff
@@ -2,26 +2,28 @@
...
-      <input
-        type="radio"
-        name="massPercentSwitch"
-        [checked]="!isMass()"
-        (change)="isMass.set(false)"
-        class="tce-switch-input" />
-      %
-    </label>
-  </span>
+  @if(shouldShowUnitsSwitch()) {
```
robatwilliams: I presume this and the other html changes in files are fixing things, rather than strictly about isolating the styles

src/app/tech-carbon-estimator/tech-carbon-estimator.component.ts
```diff
@@ -35,9 +39,40 @@ export class TechCarbonEstimatorComponent {
...
+    let stylesLink = this.createShadowStylesLink('styles.css');
+    let googleFontsLink = this.createShadowStylesLink('https://fonts.googleapis.com/icon?family=Material+Icons+Outlined');
+
+    this.ref.nativeElement.shadowRoot.appendChild(googleFontsLink);
+    this.ref.nativeElement.shadowRoot.appendChild(stylesLink);
+  }
+
+  private createShadowStylesLink(styleHref: string) {
+
+    const isAbsoluteUrl = styleHref.startsWith('http://') || styleHref.startsWith('https://');
```
robatwilliams: It may well be simpler to deal with this path wrangling in the calling method - it's only our own styles that needs anything

PR #280, opened 2025-11-13
Pdf download test

mnyamunda-scottlogic:
## Description of Change
- Test to check for PDF download
## Checklist

### DIFF
--- a/playwright-tests/e2e/T20-export-tests.spec.ts
+++ b/playwright-tests/e2e/T20-export-tests.spec.ts
@@ -6,7 +6,7 @@ import {
 } from '../utilities/test-helpers';
 import * as TestData from '../utilities/test-data';
 
-test.describe('Export JSON files', () => {
+test.describe('Download and read files', () => {
   test.beforeEach(
     async ({
       organisationSection,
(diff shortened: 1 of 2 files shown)


playwright-tests/e2e/T20-export-tests.spec.ts
```diff
@@ -79,4 +79,21 @@ test.describe('Export JSON files', () => {
...
+    let download;
+    try {
+      [download] = await Promise.all([
+        page.waitForEvent('download', { timeout: 10000 }),
+        estimationsSection.downloadPdfButton.click(),
+      ]);
+    } catch (error) {
+      throw new Error('Download failed: no file was downloaded within 10 seconds.');
+    }
+  });
```
kellis-sl: This a nice way of doing this in general JS but we cn probably make it a bit more concise for Playwright. Waiting for a download even to be triggered should be enough.

`const downloadButton = page.getByRole('button', { name: 'Download PDF' });
await Promise.all([page.waitForEvent('download'), downloadButton.click()]);`

or you could be more fancy with something like:

````const downloadButton = page.getByRole('button', { name: 'Download PDF' });
    const download = await page.waitForEvent('download');
    await downloadButton.click();
    const failure = await download.failure();
    if (failure) {
      console.log(`Download failed: ${failure}`);
    } else {
      console.log('Download succeeded');
    }
    expect(failure).toBeNull();

I'm generally against adding too much customer logging unless absolutely necessary

Comment (kellis-sl) [reply playwright-tests/e2e/T20-export-tests.spec.ts]: The other way to do this with Typescript type safety would be to create a download function (that could be reused anywhere) and then using that.

```
async function downloadFile(page: Page): Promise<Download> {
  const downloadButton = page.getByRole('button', { name: 'Download PDF' });
  const [download] = await Promise.all([
    page.waitForEvent('download'),
    downloadButton.click()
  ]);
  return download;
}


test('T20 Verify that download is executed for PDF files', async ({ page }) => {
  const download = await downloadFile(page);
  const path = await download.path();
  expect(path).toBeTruthy();
});

PR #302, opened 2025-11-26
fix: resolve all eslint errors

dmalecki-scottlogic:
Fixes all eslint errors in the project:

- Narrowed the scope of eslint checks
  - Added directories containing auto-generated files to eslint ignore section in `eslint.config.mjs`
- Allowed unused variables & function args if they're prefixed by an underscore via an addition to `eslint.config.mjs`
  - So that we can keep unimplemented functions such as `estimateDownstreamInfrastructureEmissions` without triggering eslint errors, and for test mocks that don't use an arg such as `FakeCO2Calculator`
- Replaced instances of the `any` type with explicit types where possible, `unknown` where not possible or practical
- Migrated constructor-based dependency injection to the `inject()` function
  - Firstly via the `ng generate @angular/core:inject` command
  - Fixes were then applied to broken unit tests and calling methods
    - Added `src/app/test-utils/fake-co2-calculator.testing.ts` for simplified mocking and setup without the constructor of `FakeCO2Calculator`
    - Added `TestBed` configuration and class injection to tests that previously made use of constructors
- Applied eslint automatic fixes
- Removed all unused imports, and unused variables where possible

Unit and playwright tests were run locally, and running `npx eslint .` at project root now produces zero errors.

### DIFF
--- a/src/app/assumptions-and-limitation/assumptions-and-limitation.component.spec.ts
+++ b/src/app/assumptions-and-limitation/assumptions-and-limitation.component.spec.ts
@@ -5,8 +5,7 @@ import { CarbonIntensityService } from '../services/carbon-intensity.service';
 import { WorldLocation } from '../types/carbon-estimator';
 import { gCo2ePerKwh } from '../types/units';
 
-import { CO2_CALCULATOR } from '../facades/CO2InjectionToken';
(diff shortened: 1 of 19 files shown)


lincent (approved):
looks good to me

PR #303, opened 2025-11-26
ci: add linting as pipeline check

dmalecki-scottlogic:
Adds the `lint:` step to the `buildAndTest.yml` Actions workflow file.

Lint check is run before build and test, and has been added to the `needs:` property of `build` to prevent build and test from running if there are linting errors.

See [this action run](https://github.com/ScottLogic/sl-tech-carbon-estimator/actions/runs/19709607892/job/56466296101) for an example of a linting failure in the pipeline

Currently this allows eslint warnings, but can be changed if we want to disallow warnings.

Branch protection will need to be enabled after the merge to enforce lint check success for PR merges, but this behaviour is also currently replicated by the build and test steps requiring lint checks to succeed before they run.

### DIFF
--- a/.github/workflows/buildAndTest.yml
+++ b/.github/workflows/buildAndTest.yml
@@ -4,10 +4,10 @@ on:
   workflow_dispatch:
   workflow_call:
   push:
-    # branches:
-    #   - main
-    #   - develop
-    #   - 'release/**'
+    branches:
+      - main
+      - develop
+      - 'release/**'
 jobs:
   lint:
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


.github/workflows/buildAndTest.yml
```diff
@@ -9,7 +9,17 @@ on:
...
 jobs:
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v4
+      - name: Set up environment
+        uses: ./.github/actions/set-up-environment
+      - name: Run ESLint
+        run: npx eslint .
```
lincent: I think it miught be better to run the command "npm run lint"

PR #314, opened 2025-12-11
243: Add assumptions for microsoft 365

lincent:
Add the assumptions used for the Microsoft365 emissions estimation

image


kellis-sl (approved):
Expected Playwright test failure due to screenshot comparison not matching the new text. I will resolve this when adding new tests.

dmalecki-scottlogic (approved):

PR #320, opened 2025-12-15
fix(modal): resolve host app elements obscuring export modal

dmalecki-scottlogic:
Fixes the modal being obscured by the [Tech Carbon Standard](https://www.techcarbonstandard.org/estimator) navbar, and prevents this issue from occurring in any other potential host app, by resizing the export modal to fill the space of the component only, instead of filling the entire viewport space (see screenshots).
This required additional changes:
- The export modal element was moved to the root component in the project
- The scroll behaviour of the root element was adjusted to ensure that only the export modal will scroll when it is open

### Before:
image

### After:
image

- Applied auto-formatting to src/styles.css

### DIFF
--- a/src/app/carbon-estimation/carbon-estimation.component.ts
+++ b/src/app/carbon-estimation/carbon-estimation.component.ts
@@ -21,8 +21,8 @@ import { estimatorHeights } from './carbon-estimation.constants';
 import { debounceTime, fromEvent, Subscription } from 'rxjs';
 import { CarbonEstimationTableComponent } from '../carbon-estimation-table/carbon-estimation-table.component';
 import { ExternalLinkDirective } from '../directives/external-link.directive';
-import { ExportModal } from '../export-modal/export-modal.component';
 import { CommonModule } from '@angular/common';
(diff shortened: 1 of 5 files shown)


kellis-sl (dismissed):

src/app/carbon-estimation/carbon-estimation.component.ts
```diff
@@ -63,6 +62,11 @@ export class CarbonEstimationComponent implements OnInit, OnDestroy {
 
   private changeDetectorRef = inject(ChangeDetectorRef);
 
+  @Output() public pdfExportClicked = new EventEmitter<{
```
lincent: I think it would be good to use the newer output signal here

lincent (changes requested):
look's good, just one suggestion

src/app/carbon-estimation/carbon-estimation.component.ts
lincent: https://v20.angular.dev/guide/components/outputs

lincent (approved):

PR #336, opened 2026-01-02
chore: fix npm vulnerability

kellis-sl:
Fix high risk NPM vulnerability

Update PDF download to use a consistent filename instead of the autopopulated one


lincent:
is there not an accompanying package.json file with changes? to go with the lock file?

kellis-sl:
> is there not an accompanying package.json file with changes? to go with the lock file?

I did also wonder this, but this was the only update made when running `npm audit fix`

Looking at it more closely, it's a sub dependency so not one of the packages we use directly, hence no update in the package.json

lincent:
> > is there not an accompanying package.json file with changes? to go with the lock file?
> 
> I did also wonder this, but this was the only update made when running `npm audit fix`
> 
> Looking at it more closely, it's a sub dependency so not one of the packages we use directly, hence no update in the package.json

ah I think it's because most of the versions are pre-pended with a carat `^` which means use a version that's compatible with the specified version. i.e. ^1.2.3 could have any version from 1.2.3 up to and NOT including 2.0.0.

lincent (approved):

PR #348, opened 2026-01-12
343: Fix inconsistency in form view after page reload

### LINKED ISSUE #343: [BUG] Show/Hide form behaviours do not behave as exptected on page refresh
Certain checkboxes in the form sections will hide or show additional form elements when they are checked.

The checkbox 'checked' states are retained on page refresh, but certain form elements that were previously hidden are then made visible again. 

This applies specifically to the **Cloud** and **Customer** sections of the form. Video attached.

Although these previosuly hidden elements are no longer hidden, they do not effect the calculation values, so the issue is purely a visual bug.

https://github.com/user-attachments/assets/0090b5aa-0c01-4734-a239-8d952dd69563

**Steps to reproduce the behavior:**
1. Select the 'We don't use cloud services' and 'We don't have any external users of our digital services options'
2. Confirm the relevant elements of the form are hidden.
3. Refresh page

**Expected behavior**
Previously hidden elements in the form remain hidden after page refresh

dmalecki-scottlogic:
Resolves #343 

## Description of Change

- Adds setters for form model variables to ensure form displays correctly
- Updates slider and checkbox classes to correctly show the cursor as a pointer
- Uses the default value for `estimateServerCount` signal initialisation

### DIFF
--- a/src/app/features/cloud/components/cloud-form-section.component.html
+++ b/src/app/features/cloud/components/cloud-form-section.component.html
@@ -13,7 +13,7 @@
       </div>
 
       @if (!noCloudServices()) {
-        @if (isEstimatingServers()) {
+        @if (isEstimatingServerCount()) {
           <div class="tce:flex tce:flex-col tce:gap-2">
             <label for="cloudPercentage" class="tce:text-sm"
               >What percentage of your servers are cloud services vs on-premise?</label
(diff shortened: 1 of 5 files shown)


lincent (approved):
this is probably how I would have done it. I did consider using a get method to get the value of the control and then you don't have to set up the valueChanges subscription or set the value that way. However it would mean you wouldn't be able to set the disabled state for things, such as for the downstream control

